PR #132, opened 2016-04-25
Job._get_run_times() may fall into endless loop while next_run_time equals to now

baijinping:
I made some changes.Please review.

### DIFF
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -136,9 +136,12 @@ class CronTrigger(BaseTrigger):
 
     def get_next_fire_time(self, previous_fire_time, now):
         if previous_fire_time:
-            start_date = min(now, previous_fire_time + timedelta(microseconds=1))
+            if previous_fire_time == now:
+                start_date = previous_fire_time + timedelta(microseconds=1)
+            else:
+                start_date = min(now, previous_fire_time + timedelta(microseconds=1))
         else:
-            start_date = max(now, self.start_date) if self.start_date else now
+            start_date = min(now, self.start_date) if self.start_date else now
 
         fieldnum = 0
         next_date = datetime_ceil(start_date).astimezone(self.timezone)
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -141,7 +141,7 @@ class CronTrigger(BaseTrigger):
             else:
                 start_date = min(now, previous_fire_time + timedelta(microseconds=1))
         else:
-            start_date = min(now, self.start_date) if self.start_date else now
+            start_date = max(now, self.start_date) if self.start_date else now
 
         fieldnum = 0
         next_date = datetime_ceil(start_date).astimezone(self.timezone)
--- a/tests/test_triggers.py
+++ b/tests/test_triggers.py
@@ -87,6 +87,13 @@ class TestCronTrigger(object):
         correct_next_date = timezone.localize(datetime(2016, 4, 26))
         assert trigger.get_next_fire_time(previous_fire_time, now) == correct_next_date
 
+    def test_previous_fire_time_4(self, timezone):
+        trigger = CronTrigger(day="*", timezone=timezone)
+        previous_fire_time = timezone.localize(datetime(2016, 6, 4).replace(microsecond=0))
+        now = timezone.localize(datetime(2016, 6, 4).replace(microsecond=0))
+        correct_next_date = timezone.localize(datetime(2016, 6, 5).replace(microsecond=0))
+        assert trigger.get_next_fire_time(previous_fire_time, now) == correct_next_date
+        
     def test_cron_weekday_overlap(self, timezone):
         trigger = CronTrigger(year=2009, month=1, day='6-10', day_of_week='2-4', timezone=timezone)
         assert repr(trigger) == ("<CronTrigger (year='2009', month='1', day='6-10', "
--- a/tests/test_triggers.py
+++ b/tests/test_triggers.py
@@ -82,9 +82,9 @@ class TestCronTrigger(object):
 
     def test_previous_fire_time_3(self, timezone):
         trigger = CronTrigger(day="*", timezone=timezone)
-        previous_fire_time = timezone.localize(datetime(2016, 04, 25))
-        now = timezone.localize(datetime(2016, 04, 25))
-        correct_next_date = timezone.localize(datetime(2016, 04, 26))
+        previous_fire_time = timezone.localize(datetime(2016, 4, 25))
+        now = timezone.localize(datetime(2016, 4, 25))
+        correct_next_date = timezone.localize(datetime(2016, 4, 26))
         assert trigger.get_next_fire_time(previous_fire_time, now) == correct_next_date
 
     def test_cron_weekday_overlap(self, timezone):
--- a/tests/test_triggers.py
+++ b/tests/test_triggers.py
@@ -80,6 +80,13 @@ class TestCronTrigger(object):
         correct_next_date = timezone.localize(datetime(2015, 11, 22))
         assert trigger.get_next_fire_time(previous_fire_time, now) == correct_next_date
 
+    def test_previous_fire_time_3(self, timezone):
+        trigger = CronTrigger(day="*", timezone=timezone)
+        previous_fire_time = timezone.localize(datetime(2016, 04, 25))
+        now = timezone.localize(datetime(2016, 04, 25))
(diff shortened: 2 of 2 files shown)


agronholm:
Thanks for the test case. I have quite a bit of trouble understanding the situation, but I'm definitely looking into this.

agronholm:
It seems like there are some definitions that need to be made here. What happens if the clock is set backwards? That's the only plausible case for `previous_fire_time` and `now` being exactly the same.

agronholm:
Can you rationalize your patch? It's not like `now` will stay the same value on consecutive runs. Help me understand the problem.

baijinping:
The problem is similar to the example below.
After running it, there is nothing output and CPU is busy.

---

```
import time
from datetime import datetime
from apscheduler.schedulers.background import BackgroundScheduler


scheduler = BackgroundScheduler()
scheduler.start()


def task():
    print 'run task at ', datetime.now()

trigger = CronTrigger(second='*')
scheduler.add_job(task, trigger=trigger)

time.sleep(99)
```

agronholm:
That's not what happens here. This is what I get:
```
$ python bleh.py 
run task at  2016-04-26 04:12:02.000581
run task at  2016-04-26 04:12:03.001185
run task at  2016-04-26 04:12:04.004014
run task at  2016-04-26 04:12:05.004050
run task at  2016-04-26 04:12:06.000955
run task at  2016-04-26 04:12:07.000957
run task at  2016-04-26 04:12:08.001065
run task at  2016-04-26 04:12:09.001224
run task at  2016-04-26 04:12:10.001441
run task at  2016-04-26 04:12:11.002440
run task at  2016-04-26 04:12:12.002491
run task at  2016-04-26 04:12:13.000839
run task at  2016-04-26 04:12:14.001759
run task at  2016-04-26 04:12:15.001421
run task at  2016-04-26 04:12:16.000711
run task at  2016-04-26 04:12:17.000702
run task at  2016-04-26 04:12:18.000785
```

agronholm:
By the way, when you print datetime.now(), does it not display microseconds?

baijinping:
Yes.This is what I get after I modify _get_next_fire_time():
run task at  2016-04-26 09:16:04
run task at  2016-04-26 09:16:05.001000
run task at  2016-04-26 09:16:06.002000
run task at  2016-04-26 09:16:07.001000
run task at  2016-04-26 09:16:08.002000

agronholm:
What OS and Python version are you running? You should be getting arbitrary numbers as microseconds, not something as clean as 001000 or 002000.

agronholm:
Try just running this script for a few seconds:
```python
from datetime import datetime

while True:
    print(datetime.now())
```
and tell me what it prints.

baijinping:
I run the code at Windows 7 and Python 2.7.11.

baijinping:
The output is like:
2016-04-26 09:26:31.730000
2016-04-26 09:26:31.730000
(repeat few times...)
2016-04-26 09:26:31.731000
2016-04-26 09:26:31.731000
(repeat few times...)
2016-04-26 09:26:31.732000
(repeat few times...)
2016-04-26 09:26:31.733000
(repeat few times...)

agronholm:
I see -- your computer is not capable of microsecond precision in measuring time. It can only produce millisecond precision timestamps.

agronholm:
That still doesn't explain why your script would go into an endless loop trying to compute the next run time. The code already adds one microsecond to previous_fire_time to prevent that from happening. Can you verify that adding one microsecond to a datetime works as expected?

baijinping:
Run this script will go into an endless loop in my computer.

```
import time
import pytz
from datetime import datetime
from apscheduler.schedulers.background import BackgroundScheduler


scheduler = BackgroundScheduler()
scheduler.start()


def task():
    print 'run task at ', datetime.now()

trigger = CronTrigger(second='*')
job = scheduler.add_job(task, trigger=trigger)

now = pytz.UTC.localize(datetime.now())

# block here
job._get_run_times(now)

time.sleep(99)
```

baijinping:
```
def get_next_fire_time(self, previous_fire_time, now):
        if previous_fire_time:
            start_date = min(now, previous_fire_time + timedelta(microseconds=1))
        else:
            start_date = max(now, self.start_date) if self.start_date else now
......
```
If "previous_fire_time" equals now, start_date will still be set "now" because of the "min".

agronholm:
I'll take another look at this tomorrow, but for the record, your latest script works fine here.

baijinping:
OK. I think the reason why go into an endless loop we have found -- Python can't get accurate microsecond at Windows.

agronholm:
I have a Windows machine too and it has no problem getting timestamps with microsecond accuracy.

baijinping:
How you get timestamps with microsecond accuracy on Windows?
datetime.now() or time.time()?

agronholm:
`datetime.now()` works for me.

agronholm:
The test scripts so far have reproduced the problem for you based solely on your computer hardware. But can you write a script for me that produces this infinite loop based on a fixed value of "now"? That would let me debug the issue locally.

baijinping:
I tried run datetime.now() on many WIndows machine, and got the same result(microsecond value like 345000).
I google "python datetime microsecond windows" and found some people get the same datetime.now() result like me.

baijinping:
In order to reproduce at Linux, I change apscheduler/schedulers/base.py, add some code:

```
......
from threading import RLock

from datetime import datetime as datetime_build_in
class datetime(datetime_build_in):

    @classmethod
    def now(cls, tz=None):
        now = datetime_build_in.now(tz=tz)
        old_micro = now.microsecond
        # change precision from 1000 to 1000000
        # will increase endless loop happen probability
        precision = 1000
        new_micro = old_micro - old_micro % precision
        now = now.replace(microsecond=new_micro)
        return now

from logging import getLogger
import sys
......
```

baijinping:
Hope to be helpful.

agronholm:
Hey, could you test if it makes any difference if you change line 139 in `cron/__init__.py` from `microseconds=1` to `microseconds=1000`? Also, do you get the expected result if you add `timedelta(microseconds=1)` to a datetime?

baijinping:
The problem also exists after I change ```microseconds=1``` to ```microseconds=1000```.
I got ```2016-05-05 15:57:10 127001``` after I add ```timedelta(microseconds=1)``` to ```2016-05-05 15:57:10 127000```.

agronholm:
No matter how many times I try, even with the modified datetime class you pasted, I cannot reproduce this problem.

agronholm:
The way I see it, the only way I could understand why this happens is if you give me some way to cause the infinite loop without relying on the hardware clock. Surely there is some datetime value you can directly input that will cause the infinite loop?

baijinping:
Maybe the timezone is the reason.

agronholm:
I don't see how that could be, but maybe you can try with different timezones then to confirm your theory?

baijinping:
When scheduler add a job, the job's attribute "next_run_time" will set to now.
```
# Calculate the next run time if there is none defined
if not hasattr(job, 'next_run_time'):
    now = datetime.now(self.timezone)
    replacements['next_run_time'] = job.trigger.get_next_fire_time(None, now)
```

And the microsecond value in job.next_run_time equals to zero(after ```datetime_ceil()```).

Then call ```_get_run_times(now)```, which now.microsecond may equals to zero.

If there is less then one second between call ```_real_add_job``` and ```_get_run_times```, the arguments sent into ```_get_next_fire_time``` will be equals(previous_fire_time == now).

In this situation, the code below can't help.
```
if previous_fire_time:
    start_date = min(now, previous_fire_time + timedelta(microseconds=1))
else:
    start_date = max(now, self.start_date) if self.start_date else now
```

Then return the same date time value.
The endless loop happen now.

agronholm:
Your logic is a bit flawed. If `now` and `previous_fire_time` are exactly the same, `start_date` would end up `now` + 1 microsecond.

agronholm:
I also don't understand why the microsecond being zero would have any significance.

agronholm:
Can you construct a Job and a datetime value for `now` in such a way that calling job._get_run_times(now) would end up in an infinite loop?

agronholm:
Never mind, I managed to do that on my own now.

baijinping:
The new test may show the problem in you computer.

agronholm:
I have fixed this in 21aa9a944247f4e05dfc91fc8dfed23b5d2b6d89 and credited you in the version history. I did not directly use your patch except for one of the tests. Switching from max() to min() in the case when previous_fire_time is `None` would have caused some fire times to be missed.

PR #134, opened 2016-04-28
Trying to fix #96

### LINKED ISSUE #96: AsyncioExecutor not truly async
Originally reported by: **Lennart Bastian (Bitbucket: [bastianl](http://bitbucket.org/bastianl), GitHub: [bastianl](http://github.com/bastianl))**

----------------------------------------

I have been using the `AsyncioScheduler` with the `APScheduler` to schedule tasks asynchronously. However, I want to schedule jobs (coroutines) in the same event loop that the scheduler is running in. 
It looks like the `AsyncioExectutor` is running tasks in a [separate thread](https://bitbucket.org/agronholm/apscheduler/src/f312c39b161011b6c6372960b6985434854fab84/apscheduler/executors/asyncio.py?at=master#cl-27), and not as coroutines but rather regular functions. In fact, it doesn't seem like you can schedule any kind of asynchronous functions. Is this the intended behavior? What's the point of calling it an `asyncio` executor? 

To solve this I modified the function `run_job` to work asynchronously, and tweaked the executor to run jobs as `asyncio.Tasks` which are automatically inserted into the eventloop.
[See this](https://gist.github.com/bastianl/e7a1879f583e69f95c0e).

Maybe we can modularize the `run_job` function and split it into two functions, one of which works asynchronously. Let me know what you think, am happy to submit a pull request.

----------------------------------------
- Bitbucket: https://bitbucket.org/agronholm/apscheduler/issue/96

txomon:
So, @agronholm it took me one day of work to figure out the best layout for asyncio.

I did a lot of experiments, but this I think is the best answer. I haven't looked too much for python3 backwards compatibility, but it should support python3.3 onwards.

Asyncio after all is not available before (AFAIK)

### DIFF
--- a/apscheduler/executors/asyncio.py
+++ b/apscheduler/executors/asyncio.py
@@ -1,7 +1,11 @@
 from __future__ import absolute_import
-import sys
 
-from apscheduler.executors.base import BaseExecutor, run_job
+import asyncio
+import sys
+from traceback import format_tb
+
+from apscheduler.events import JobExecutionEvent, EVENT_JOB_ERROR, EVENT_JOB_EXECUTED
+from apscheduler.executors.base import BaseExecutor
 
 
 class AsyncIOExecutor(BaseExecutor):
@@ -16,6 +20,9 @@ class AsyncIOExecutor(BaseExecutor):
         self._eventloop = scheduler._eventloop
 
     def _do_submit_job(self, job, run_times):
+        asyncio.get_event_loop().call_soon(self._do_job_runtime, job, run_times)
+
+    def _do_job_runtime(self, job, run_times):
         def callback(f):
             try:
                 events = f.result()
@@ -24,6 +31,29 @@ class AsyncIOExecutor(BaseExecutor):
             else:
                 self._run_job_success(job.id, events)
 
-        f = self._eventloop.run_in_executor(None, run_job, job, job._jobstore_alias, run_times,
-                                            self._logger.name)
-        f.add_done_callback(callback)
+        events = self._job_runtime(job, run_times)
+        future_events = []
+        for event in events:
+            if not (isinstance(event, asyncio.Future) or asyncio.iscoroutine(event)):
+                future = asyncio.Future()
+                future.set_result(event)
+                future_events.append(future)
+            else:
+                future_events.append(event)
+        future = asyncio.gather(*events)
+        future.add_done_callback(callback)
+
+    @asyncio.coroutine
+    def _run_job(self, job, run_time):
+        """Actual implementation of calling the job function"""
+        try:
+            retval = yield from job.func(*job.args, **job.kwargs)
+        except:
+            exc, tb = sys.exc_info()[1:]
+            formatted_tb = ''.join(format_tb(tb))
+            self._logger.exception('Job "%s" raised an exception', job)
+            return JobExecutionEvent(EVENT_JOB_ERROR, job.id, job._jobstore_alias, run_time,
+                                     exception=exc, traceback=formatted_tb)
+        else:
+            self._logger.info('Job "%s" executed successfully', job)
+            return JobExecutionEvent(EVENT_JOB_EXECUTED, job.id, job._jobstore_alias, run_time, retval=retval)
--- a/apscheduler/executors/asyncio.py
+++ b/apscheduler/executors/asyncio.py
@@ -44,9 +44,10 @@ class AsyncIOExecutor(BaseExecutor):
                 self._run_job_success(job.id, events)
 
         if not asyncio.iscoroutinefunction(job.func):
-            future = self._eventloop.run_in_executor(None, job_runtime, job, run_times, self._logger.name)
+            future = self._eventloop.run_in_executor(None, job_runtime, job, run_times, job._jobstore_alias,
+                                                     self._logger.name)
         else:
-            events = job_runtime(job, run_times, self._logger.name, self._run_job)
+            events = job_runtime(job, run_times, self._logger.name, job._jobstore_alias, self._run_job)
             future_events = []
             for event in events:
                 if not (isinstance(event, asyncio.Future) or asyncio.iscoroutine(event)):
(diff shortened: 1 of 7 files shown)


txomon:
BTW, I expect tests to fail terribly because I didn't touch them...

agronholm:
I see quite a few problems here:

1. This will break Python 2.7 compatibility due to the use of "yield from" and not supporting the Trollius backport (which AsyncIOScheduler *does*)
1. This will break backwards compatibility because jobs AsyncIOExecutor now always runs everything in the event loop thread where it previously ran jobs in the default (thread pool) executor
1. BasePoolExecutor submits a task pointing to a bound method, which will break process pool based executors (this is the reason `run_job()` was a free function)
1. It duplicates a lot of code, which I was specifically hoping to avoid

agronholm:
I don't want to discourage you, but this is a hard problem and may require one or more backwards incompatible API changes to be properly implemented.

txomon:
Thanks for the fast review!

  1. I completely agree, I had totally forgotten about trollius =) I will make the changes to support 2.7
  2. This was done on purpose. With asyncio, we should not need more than one thread. This is my main complaint about the current implementation and the reason for this patch
  3. I will have a look on that
  4. Only duplicates the necessary parts in code to support both coroutines and normal functions. If you check the diff, I have only added the necessary support for Futures, etc.

About 1, I totally forgot about trollius, so breaking 2.7 compatibility wasn't a problem because asyncio couldn't be used.

agronholm:
Regarding #2, a point release should not break backwards compatibility (although this inadvertently happened with 3.1). However, as coroutine callables didn't work to begin with, it'd be acceptable to get them to work while still running regular functions in the default thread pool.

txomon:
BTW, reading the trollius documentation, it is a deprecated project... http://trollius.readthedocs.io/deprecated.html

agronholm:
I'm aware. But it does work.

agronholm:
In APScheduler 4.0 I will think about ditching Trollius.

apscheduler/executors/asyncio.py
```diff
@@ -31,29 +43,54 @@ def callback(f):
...
-        except:
-            exc, tb = sys.exc_info()[1:]
-            formatted_tb = ''.join(format_tb(tb))
-            self._logger.exception('Job "%s" raised an exception', job)
-            return JobExecutionEvent(EVENT_JOB_ERROR, job.id, job._jobstore_alias, run_time,
-                                     exception=exc, traceback=formatted_tb)
-        else:
-            self._logger.info('Job "%s" executed successfully', job)
-            return JobExecutionEvent(EVENT_JOB_EXECUTED, job.id, job._jobstore_alias, run_time, retval=retval)
+    if real_asyncio:
```
agronholm: I hope you're not being serious about this duplication. It also won't fix the syntax error. There are better ways to emulate "yield from" in a way that works for both asyncio and trollius.

txomon:
@agronholm Perfect, nice to see proper semantic versioning =)

Anyway, I have updated this PR with the latest solution, re-extracting the run_job to make pool executors happy, and created a dirty if/else for trollius.From / trollius.Return

Is there anything else, apart from that that I should take into account?

apscheduler/executors/asyncio.py
txomon: I am all ears =)

The alternative was to place if else depeding on trollius or asyncio for the returns and the yield from, but that looked better. 

I am really out of ideas on how to emulate yield from.

agronholm: Native coroutines have the `__await__` method which returns an iterable. On Python < 3.4, you can just iterate over the generator directly:
```python
for v in job.func(*job.args, **job.kwargs):
    retval = yield v
```
I haven't actually done this kind of emulation before but this should help you figure out a common code path that works for both asyncio and trollius plus all versions of Python.

agronholm: There is also an option of using the ``async()`` or ``ensure_future()`` function which should work nicely together with ``run_on_executor`` which also returns a future. You'll need to experiment a bit though!

txomon: > Native coroutines have the __await__ method which returns an iterable. On Python < 3.4, you can just iterate over the generator directly:

The problem is that they don't accept return inside generators, I already tried and the function doesn't get to execute, that's why trollius has the raise Return()

> There is also an option of using the async() or ensure_future() function which should work nicely together with run_on_executor which also returns a future. You'll need to experiment a bit though!

The problem with these is that you need someone to read the future result, or it won't execute. ATM, with the future + callback it works smoothly

agronholm:
A "return" in a generator function is still a syntax error in Python 2, so this won't work. But I like your notion of moving run_job() to the executor class. It just needs to be a class method or static method in order to work with PoolExecutor (using process pools). This just needs a lot more deduplication and some innovative refactoring.

txomon:
@agronholm But the `if real_asyncio:` is taking care of that, isn't it? That's the whole point of using raise() instead of return in the else clause...

agronholm:
No, you don't understand. Merely importing the module causes a SyntaxError in 2.7. Try it.

txomon:
@agronholm, I need to add tests then. The only alternative I can think of really is separating it in two files... It's the cleanest solution, with no dirty if/else, we would need to duplicate a little code, but I think it's the best solution, does any alternative come to your mind?

agronholm:
I'm still banking on using `create_task()` or `ensure_future()` and adding a callback for that. Unfortunately there are higher priority issues for me to fix so I can't really concentrate on this feature yet.

agronholm:
I've taken another look of this problem in depth. The code that checks for misfires must absolutely run in the same task as the target function, so "yield from" is a must. It's starting to look like code duplication is pretty much unavoidable, at least until the next major version where the API can be renovated. The immediate problem here is using "yield from" in a way that still lets the library work on Python 2.

txomon:
Yes, the only way I found was to have two modules, one for Python with
tulip and the other for Python3 with coroutines

agronholm:
I ended up limiting support to native coroutines and as such, excluding trollius. Even then it was fairly painful. Thank you for your effort even though it didn't end up merged.

txomon:
Yeah, it was really painful for me too, don't worry about the code!! Thanks
for the hard work!

PR #136, opened 2016-04-29
First implementation of the Google Datastore Store for APScheduler

txomon:
So my company is using google appengine and I have implemented a job store that uses the google datastore.

I haven't done too deep testing, but all the basic methods are implemented, and because I am serializing everything, there shouldn't be any problem with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ docs/_build/
 build/
 virtualenv/
 example.sqlite
+/.eggs/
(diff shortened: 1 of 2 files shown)


agronholm:
Just so you know I don't accept ANY new code that doesn't add tests to cover it.

txomon:
@agronholm sure, I just need to figure out how to setup the testing env in travis, and how all the fixtures etc. stick together

agronholm:
This has stalled for 1½ years. If you want to try again, open a new PR.

PR #137, opened 2016-04-29
Easy fix of exception returning id function instead of job id

txomon:
I have no idea on how to test this thing. I suppose it would mean modifying test_jobstores:test_update_job_nonexistent_job() but I don't know how to test that the attribute we complain of is 'blah' and not any other.


agronholm:
Ah, now I understand the problem. This should be easy to test by checking the exception message. I can do that myself though. Thank you.

PR #144, opened 2016-05-24
Zookeeper jobstore support

jvillar:
I'm using Zookeeper for HA coordination of distributed systems and I've implemented a job store for Apscheduler that uses Zookeeper as storage backend.

I've implemented tests and everything seems Ok, with a coverage of 98% for python 2.7 and python 3.4 and all tests passed.

**Tox output:**

> apscheduler/jobstores/zookeeper.py           118      2    98%   127-128
> py27: commands succeeded
> py34: commands succeeded

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,10 @@ python:
 before_install:
   - docker pull mongo:latest
   - docker pull rethinkdb:latest
+  - docker pull jplock/zookeeper
   - docker run -d -p 127.0.0.1:27017:27017 mongo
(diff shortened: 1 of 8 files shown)


agronholm:
Thanks, looks very professional work! I will merge this soonish. Meanwhile, please check why the Travis run failed.

jvillar:
Hi!
the problem has already been fixed. When asking for the children of a zookeeper node, Kazoo does not  guarantee any order but it was returning an ordered list on my local tests. The order of the entries is now explicitly defined by the zookeeper jobstore class.

PR #154, opened 2016-08-09
CronTrigger behavior around DST transitions

backbord:
Hi @agronholm ,

I have been working on making the CronTrigger respect daylight saving time transitions and would gladly like to share this functionality with you.

Although I'm unsure if I managed to match your coding style, I kindly ask for a review of this PR.
Please let me know if it can be improved and if you would be willing to accept it.

I tried to ensure that the changes are compatible with your documentation (http://apscheduler.readthedocs.io/en/latest/modules/triggers/cron.html#daylight-saving-time-behavior) in the sense that they narrow down the behavior to "execute exactly when the cron trigger expression matches the wall clock in the time zone -- even considering daylight saving time."

Best regards and keep up the good work,
Tim

### DIFF
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -180,13 +180,13 @@ class CronTrigger(BaseTrigger):
         # check if a switch from no dst to dst falls into the 24 hours before the next date
         if next_date \
                 and next_date.dst() > self.timezone.normalize(next_date - timedelta(days=1)).dst():
-            # perform a fixed point iteration to cope with "missing time" 
+            # perform a fixed point iteration to cope with "missing time"
             # due to a switch from winter to summer time
             last_date = None
             loops = 0
             while last_date != next_date:
                 if loops > 10 and next_date - start_date > timedelta(days=5*365):
-                    # if there's no valid fire time in the next ~5 years, 
+                    # if there's no valid fire time in the next ~5 years,
                     # then give up and return "no next fire time"
                     return None
 
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -133,7 +133,7 @@ class CronTrigger(BaseTrigger):
                     values[field.name] = new_value
 
         return self.timezone.normalize(self.timezone.localize(datetime(**values),
-                                       is_dst=dateval.dst()))
+                                                              is_dst=dateval.dst()))
 
     def _find_next_potential_fire_time(self, start_date):
         next_date = start_date
@@ -178,8 +178,8 @@ class CronTrigger(BaseTrigger):
         next_date = self._find_next_potential_fire_time(start_date)
 
         # check if a switch from no dst to dst falls into the 24 hours before the next date
-        if next_date \
-                and next_date.dst() > self.timezone.normalize(next_date - timedelta(days=1)).dst():
+        if (next_date and
+                next_date.dst() > self.timezone.normalize(next_date - timedelta(days=1)).dst()):
             # perform a fixed point iteration to cope with "missing time"
             # due to a switch from winter to summer time
             last_date = None
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -132,16 +132,11 @@ class CronTrigger(BaseTrigger):
                 else:
                     values[field.name] = new_value
 
-        return self.timezone.localize(datetime(**values))
+        return self.timezone.normalize(self.timezone.localize(datetime(**values),
+                                       is_dst=dateval.dst()))
 
-    def get_next_fire_time(self, previous_fire_time, now):
-        if previous_fire_time:
-            start_date = min(now, previous_fire_time + timedelta(microseconds=1))
-            if start_date == previous_fire_time:
-                start_date += timedelta(microseconds=1)
-        else:
-            start_date = max(now, self.start_date) if self.start_date else now
-
+    def _find_next_potential_fire_time(self, start_date):
+        next_date = start_date
         fieldnum = 0
         next_date = datetime_ceil(start_date).astimezone(self.timezone)
         while 0 <= fieldnum < len(self.fields):
@@ -169,6 +164,38 @@ class CronTrigger(BaseTrigger):
 
         if fieldnum >= 0:
             return next_date
+
+    def get_next_fire_time(self, previous_fire_time, now):
+        if previous_fire_time:
+            start_date = min(now, previous_fire_time + timedelta(microseconds=1))
+            if start_date == previous_fire_time:
+                start_date += timedelta(microseconds=1)
+        else:
+            start_date = max(now, self.start_date) if self.start_date else now
+
+        start_date = datetime_ceil(start_date).astimezone(self.timezone)
+
+        next_date = self._find_next_potential_fire_time(start_date)
+
+        # check if a switch from no dst to dst falls into the 24 hours before the next date
+        if next_date \
+                and next_date.dst() > self.timezone.normalize(next_date - timedelta(days=1)).dst():
+            # perform a fixed point iteration to cope with "missing time" 
+            # due to a switch from winter to summer time
+            last_date = None
+            loops = 0
+            while last_date != next_date:
+                if loops > 10 and next_date - start_date > timedelta(days=5*365):
+                    # if there's no valid fire time in the next ~5 years, 
+                    # then give up and return "no next fire time"
+                    return None
+
+                last_date = next_date
+                loops += 1
+
+                next_date = self._find_next_potential_fire_time(last_date)
+
+        return next_date
 
     def __getstate__(self):
         return {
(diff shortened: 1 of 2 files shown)


agronholm:
Any reason you haven't used the `timezone` fixture like the previous tests do?

What is the gist of this change? I mean, what changes here?

backbord:
Not using the `timezone` fixture in `test_dst_missing_hour` was a left-over from a more extensive test suite where I intended to emphasize the behavior in different timezones. For the PR I just changed it to use the fixture.

The gist of `test_dst_missing_hour` is, that on the last sunday of March in time zone Europe/Berlin, the wall clock is set forward from 2 to 3 o'clock, effectively missing the hour from 2 to 3.

If a trigger is scheduled to only ever execute in such a missing hour, it must never be executed.
(However, note that the time zone handling provided by `pytz` appears to only work until the the end of the unix epoch in 2038.)

The `test_dst_change` cases check several cases of CronTrigger's behavior in the context of DST transitions.

backbord:
The gist of the change is:

A CronTrigger with a time zone that has DST transitions will now fire if and only if the a clock (in that time zone) shows a matching time.
This results in
- triggers being executed twice for the "same time" (w.r.t. the time zone) when a time frame is repeated during a transition from DST to non-DST and
- triggers not being executed for times that do not exist (w.r.t. the time zone) when a time frame is "skipped" during a transition from non-DST to DST.
Note that triggers that are configured to only ever execute in a "skipped" time frame will never be executed.

The change comprises of several parts:
- _set_field_value normalizes the returned datetime and explicitly sets the DST value to that of the input dateval.
- _find_next_potential_fire_time(start_date) has the property to return the start_date if it is a valid datetime for the time zone and the next potential fire time if the start_date does not exist in the time zone.
- get_next_fire_time uses _find_next_potential_fire_time to find the next potential fire time which is checked against "skipped" time frames during a transition from non-DST to DST and performs a fixed point search for the next valid fire time. The fixed point search is limited to roughly 5 years.

(comment for https://github.com/agronholm/apscheduler/pull/154/commits/ec7969f80980551aa529be873cbe4f9bb826a816)

agronholm:
I've been rewriting parts of the cron scheduling code for my [asphalt-tasks](https://github.com/asphalt-framework/asphalt-tasks) project. A couple of cron related fixes for bugs discovered by the rewritten test suite were even recently backported to APScheduler. I already have passing tests in place that correctly work with DST shifts but I'll have to run your tests against the code too. Frankly I prefer to use my own code as yours adds a considerably large number of lines for something this small.

backbord:
I'm happy to check out [asphalt-tasks](https://github.com/asphalt-framework/asphalt-tasks), but the repo looks empty to me.

agronholm:
I just pushed all my committed local code to the repository.

backbord:
Thanks.

agronholm:
I just adapted your test_dst_change() tests to asphalt-tasks. 3 failures. Will investigate.

agronholm:
OK, the autumn DST crossover date in your tests is wrong for the Europe/Berlin timezone. The date in your tests is 27th October while the actual crossover date is 30th October. How did you get the tests to pass locally?

backbord:
Hi, the `test_triggers.py` code has `DST_AUTUMN = datetime(2015, 10, 25, 2)` for `DST_TIMEZONE = 'Europe/Berlin'` which is used in `test_dst_change`.

If I CTRL+F for `27`, I get a single occurrence: `DST_SPRING = datetime(2016, 3, 27, 2)`.

While it might be irritating that I chose the last two transitions as they fall into different years, both dates are correct transitions: https://www.timeanddate.com/time/zone/germany/berlin

If you prefer both dates to fall into the same year, I'm happy to change the dates for the tests.

Can you point me to the precise line number where you see October 27th?

Note that `test_dst_missing_hour` does not employ a fixed date for the trigger but CronTriggers very nice syntax for "the last Sunday of March" (thank you for that functionality! love it!) as that is the day on which the spring DST transition happens in Europe/Berlin. (Apologies if I am stating the obvious here...)

agronholm:
My bad, I didn't consider the year. Sorry. I'll keep investigating.

backbord:
Ok. I don't know if it helps, but before the CI not wanting to build the PR code due to an error, the tests went through for this python2.7 build https://travis-ci.org/agronholm/apscheduler/jobs/150929738 -- unfortunately I overlooked an old print statement which led to failed tests for python3.

By the way, do you know how to repair the Travis CI build system?
It could be that a mongo instance is already running and bound to port 27017.
But I don't know how I can solve the issue -- or if I'm even allowed to.
`$ docker run -d -p 127.0.0.1:27017:27017 mongo
cd7a03d772453b4c6c85bf40d8c8f34b14d29f7b399f0b5bfcaab6d2135eb53d
docker: Error response from daemon: driver failed programming external connectivity on endpoint adoring_goldwasser (8aec5565975d75669534d5d50cfd60e2cbc708b92289b295b33223758520ecac): Error starting userland proxy: listen tcp 127.0.0.1:27017: bind: address already in use.`
(From https://travis-ci.org/agronholm/apscheduler/jobs/151138242)

agronholm:
The reason the builds are failing is this: https://www.traviscistatus.com/incidents/2p40l49r3yxd
Apparently an unwanted mongodb instance is being launched on container boot (this I was told by Travis support). They know about the problem and are fixing it.

agronholm:
I solved the problem with DST backward shifts. Only one forward shift test is failing now.

agronholm:
The "time gap" from the forward shift is giving me a headache. Makes me want to rewrite most of the CronTrigger code.

backbord:
I agree that the forward shift is not as nice to deal with as the backward shift.
In the PR, I wanted to change as little code as possible (w.r.t. my understanding of your code and the problem), but I did not find a clever way to avoid the iteration.

agronholm:
I suggest not spending much effort on this as I will probably be rewriting the code myself.

backbord:
Yeah, glad to read that. Your work is greatly appreciated! Thanks.

backbord:
Hi, any news on this topic?

agronholm:
Nope, haven't gotten around to finishing the work. I will get to it eventually.

backbord:
Hi, any news?
The last couple of DST transitions went well for our systems which run patched version of APScheduler and I would love to have this feature in the official version. :-)
Best regards,
Tim

agronholm:
Oh dear, it's already been a whole year. But trust me, I have not forgotten about this. I will implement the change first in my [asphalt-scheduler](https://github.com/asphalt-framework/asphalt-scheduler) project (which is direly needed!) and then I will backport it here. I've been super busy all around (I took up maintainership of the [wheel](https://github.com/pypa/wheel) project lately), that's one reason why this hasn't happened yet. Sorry about it.

agronholm:
I've resumed work on this, just letting you know.

agronholm:
I've pushed my own version of the fix to the cron-dst-fix branch. I would appreciate if you reviewed the changes for me.

backbord:
Hi Alex,

aside from the aesthetics of your code (clean and concise as always) I found that all but one of my tests passed nicely.

The only miniscule "issue" I found was the rather obscure edge case where a trigger is defined that should not fire as the specified time "02:30 on the last sunday of march each year" always falls into the forward transition of Europe/Berlin.
```
    def test_dst_missing_hour(self, timezone):
        """
        It might be that a CronTrigger can be configured to only ever fire in the missing
        time period during the switch from standard to daylight saving time,
        resulting in a trigger unable to fire.

        """
        trigger = CronTrigger(timezone=timezone, month=3, day='last sun', hour=2, minute=30)
        start_date = timezone.localize(datetime(2006, 4, 1), is_dst=False)
        next_date = trigger.get_next_fire_time(None, start_date)
        assert next_date is None
```
Here, I would expect `next_date` to be `None` but the code gives a timestamp after the end of the unix epoch (2038-01-19T3:14:08 UTC), namely `datetime.datetime(2038, 3, 28, 2, 30, tzinfo=<DstTzInfo 'Europe/Berlin' CET+1:00:00 STD>)`. This issue does not stem from the code in apscheduler (my PR does not cope with it either).
This is very much "good enough" in my opinion. :)
I guess I'd consider including a test like the one above for that obscure edge case (as a documentation of this behavior) in the test suite but with `assert next_date is None or next_date > datetime(2038, 1, 19)` -- "all bets are off after the end of the epoch."

Thank you very much for your well-written and clean solution! Looking forward to seeing that in an upcoming release!

Have a nice weekend,
Tim

agronholm:
Thanks for the review! I was surprised to learn that you can get it to fail in a way like this. It will be interesting to figure out what causes this.

This fix will have to wait until the 4.0.0 release which is ways off, because it changes DST semantics from what users expected before, even if they were wrong. Meanwhile, I am applying this to the [asphalt-scheduler](https://github.com/asphalt-framework/asphalt-scheduler) project (code to be pushed into the repository very soon). This is a reimagining of APScheduler in the context of the Asphalt framework, with full integration and much better asyncio support.

backbord:
You're welcome!
The issue stems from the zoneinfo database of pytz which lists transition times only until 2038 (see https://github.com/newvem/pytz/blob/master/pytz/zoneinfo/Europe/Berlin.py#L155 for example -- this seems to be a copy of the pytz package, I wasn't able to fully identify the code that generates the zoneinfo files in the pytz repo https://git.launchpad.net/pytz/tree/).
I do believe that to be way out of the scope of apscheduler.

agronholm:
Yes, I only now saw this comment of yours and I had just come to the same conclusion. Thanks again!

PR #166, opened 2016-11-10
Track lifecycle & history of jobs by storing `job_submissions` in jobstore

seanharr11:
# Job Submissions

This is a large change that essentially inserts a `job_submission` into the jobstore each time a job is run by the executor, and tracks this `job_submission` throughout its lifecycle. A `job_submission` object is defined as follows:

> **int|str id**: _Unique identifier, int|str depending on the jobstore/strategy for UID generation_
> **int|str apscheduler_job_id**: _Identifier which references the `job` which spawned this `job_submission`_
> **str func:** _The **name** of the function which the job is scheduled to run. This is preserved on the job_submission so that if the `job` is removed from the jobstore, we can still run analytics on job_submissions based on the python function that this submission ran._
> **datetime submitted_at**: _The date & time the job_submission was submitted by the executor._
> **datetime completed_at**: _The date & time that the executor reported the job as completing._
> **str state**:  _The state that the job_submission is currently in._
> - _submitted_ - The job has been submitted by the scheduler, to the executor to be run
> - _success_ - The job has completed successfully
> - _failure_ - The job has completed unsuccessfully
> - _missed_ - The job was missed at the given `submitted_at` time.
> - _orphaned_ - The job was previously in a _submitted_ state, but the scheduler or jobstore ungracefully stopped without updating the job, leaving it running or stopped in an untracked state.
> 

Other changes include:

1. Executors are no longer responsible for running a job several times, depending on the `run_times` argument passed to `submit_job`. This responsibility now falls on the `scheduler`. If job.coalesce = False, and the missed run times fall within the `job.grace_period`, the scheduler will call `submit_job()` for each run_time when it is looking through jobs in `process_jobs()`.
2. The following public methods have been added to each jobstore to support job_submissions:

   - `add_job_submission(self, job, now)`
   - `update_job_submission(self, job_submission_id, **kwargs)`
   - `update_job_submissions(self, conditions, **kwargs)`
   - `get_job_submissions_with_states(self, states=[])`
   - `get_job_submission(self, job_submission_id)`
_* Note that no method for DELETE is implemented for CRUD surrounding job_submissions, as a job_submission should never be deleted.

3. The Scheduler's Public API has gained 3 methods, providing READ access to job_submissions:
   - `get_job_submissions(self, jobstore)`
   - `get_job_submission(self, jobstore, job_submission_id)`
   - `get_job_submissions_for_job(self, jobstore, job_id)`
4. JobExecutionEvents now take a `job_submission_id` and a `scheduled_run_time` has parameters to support job_submission listeners via Public Events API.

All tests are passing via tox (on all Python distributions), all above functions have added unit tests that are passing, and flake8 tests pass concerning PEP8 compliance.

### DIFF
--- a/apscheduler/events.py
+++ b/apscheduler/events.py
@@ -85,8 +85,8 @@ class JobExecutionEvent(JobEvent):
     :ivar traceback: a formatted traceback for the exception
     """
 
-    def __init__(self, code, job_id, jobstore, scheduled_run_time, job_submission_id=None, retval=None, exception=None,
-                 traceback=None):
+    def __init__(self, code, job_id, jobstore, scheduled_run_time,
+                 job_submission_id=None, retval=None, exception=None, traceback=None):
         super(JobExecutionEvent, self).__init__(code, job_id, jobstore)
         self.scheduled_run_time = scheduled_run_time
         self.retval = retval
--- a/apscheduler/events.py
+++ b/apscheduler/events.py
@@ -85,10 +85,11 @@ class JobExecutionEvent(JobEvent):
     :ivar traceback: a formatted traceback for the exception
     """
 
-    def __init__(self, code, job_id, jobstore, scheduled_run_time, retval=None, exception=None,
+    def __init__(self, code, job_id, jobstore, scheduled_run_time, job_submission_id=None, retval=None, exception=None,
                  traceback=None):
         super(JobExecutionEvent, self).__init__(code, job_id, jobstore)
         self.scheduled_run_time = scheduled_run_time
         self.retval = retval
         self.exception = exception
         self.traceback = traceback
+        self.job_submission_id = job_submission_id
--- a/apscheduler/executors/asyncio.py
+++ b/apscheduler/executors/asyncio.py
@@ -29,12 +29,8 @@ class AsyncIOExecutor(BaseExecutor):
 
     def _do_submit_job(self, job, job_submission_id, run_time):
         def callback(f):
-            try:
-                events = f.result()
-            except:
-                self._run_job_error(job.id, job_submission_id, job._jobstore_alias, *sys.exc_info()[1:])
-            else:
-                self._run_job_success(job.id, job_submission_id, job._jobstore_alias, events)
+            event = f.result()
+            self._handle_job_event(event)
 
         if iscoroutinefunction(job.func):
             if run_coroutine_job is not None:
--- a/apscheduler/executors/asyncio.py
+++ b/apscheduler/executors/asyncio.py
@@ -6,7 +6,6 @@ from apscheduler.executors.base import BaseExecutor
 
 try:
     from asyncio import iscoroutinefunction
-    from apscheduler.executors.base_py3 import run_coroutine_job
 except ImportError:
     from trollius import iscoroutinefunction
     run_coroutine_job = None
--- a/apscheduler/executors/asyncio.py
+++ b/apscheduler/executors/asyncio.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 
 import sys
 
-from apscheduler.executors.base import BaseExecutor, run_job
+from apscheduler.executors.base import BaseExecutor
 
 try:
     from asyncio import iscoroutinefunction
@@ -27,7 +27,7 @@ class AsyncIOExecutor(BaseExecutor):
         super(AsyncIOExecutor, self).start(scheduler, alias)
         self._eventloop = scheduler._eventloop
 
-    def _do_submit_job(self, job, run_times):
+    def _do_submit_job(self, job, run_times, run_job_func):
         def callback(f):
             try:
                 events = f.result()
@@ -43,7 +43,7 @@ class AsyncIOExecutor(BaseExecutor):
             else:
                 raise Exception('Executing coroutine based jobs is not supported with Trollius')
         else:
-            f = self._eventloop.run_in_executor(None, run_job, job, job._jobstore_alias, run_times,
+            f = self._eventloop.run_in_executor(None, run_job_func, job, job._jobstore_alias, run_times,
                                                 self._logger.name)
 
         f.add_done_callback(callback)
--- a/apscheduler/executors/asyncio.py
+++ b/apscheduler/executors/asyncio.py
@@ -32,18 +32,18 @@ class AsyncIOExecutor(BaseExecutor):
             try:
                 events = f.result()
             except:
-                self._run_job_error(job.id, *sys.exc_info()[1:])
+                self._run_job_error(job.id, job_submission_id, job._jobstore_alias, *sys.exc_info()[1:])
             else:
-                self._run_job_success(job.id, events)
+                self._run_job_success(job.id, job_submission_id, job._jobstore_alias, events)
 
         if iscoroutinefunction(job.func):
             if run_coroutine_job is not None:
-                coro = run_coroutine_job(job, self._logger.name, job_submission_id, run_time)
+                coro = run_coroutine_job(job, self._logger.name, job_submission_id, job._jobstore_alias, run_time)
                 f = self._eventloop.create_task(coro)
             else:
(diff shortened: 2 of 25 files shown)


agronholm:
Quite a large PR. Seems pretty risky doing all this work without discussing it with me first. But now that you've submitted it, I'd like to hear the motivations behind these changes, in as much detail as you can provide.

seanharr11:
Perhaps I got ahead of myself not discussing with you first, but I would have made these changes either way to accomplish the needs of my team: I run a Software team for the New England Patriots, and chose APScheduler to run all sorts of Machine Learning and Analytics Jobs for the coaching staff, scouting department, training staff, and IT department.

What it boils down to is that we needed a way to track jobs throughout their lifecycle, without relying on additional dependencies like ElasticSearch, or parsing logs. Different insights we need include:

1. Get a list of all jobs running at a given time
    - This can be later expanded to allow for 'stop', 'kill' or other system-level commands to be sent to jobs.
2. Get the historical average run time of a job
   - This is crucial in the high-pressure environment we work in. Before a coach/scout runs a job, we need to let them know how long it will take.
   - We also need to know when jobs take an uncharacteristically long time, as this may point to a resource bottleneck.
   - We can also "predict" how far along a given job is based on its history, allowing end-users to build progress bars in their UIs.
3. Get a specific job's failed executions (primarily so we can reduce failure rate, and have a way to quantify it)
4. Identify job submissions with statuses of 'orphaned' or 'missed', so we can identify these instead of them slipping under the covers (for instance if the scheduler crashes ungracefully)
5. Add additional granularity to the Events API. With the `job_submission_id` stored on the JOB_EXECUTION_EVENT and JOB_FAILED_EVENT, end-users have more context in their callback functions, allowing them to get the job_submission from the jobstore, and take action with this data (the job_submission object will likely gain metadata fields, and users will want these available in their callbacks)

My team plans on building an Angular2 dashboard to view this data, which will talk to Flask-APScheduler. I have already readied a pull-request for Flask-APScheduler to support READ operations to get the above job_submission data. This PR really provides end-users with the data to view the overall state of our APScheduler system, and gives us the insights to see how well our jobs are running.

My guess is that making these insights available out-of-the-box will only increase the adoption/contributions to APScheduler, but I understand that you've got to control scope and features. I would certainly volunteer myself to maintain this part of the project as you add features.

Please let me know your thoughts.

seanharr11:
@agronholm any thoughts here? Could hop on a video chat if you'd like.

agronholm:
Throughout the 7 year history of APScheduler, you are the first to ask for job history retention. There are many ways to do it and I'm not convinced at the moment that its inclusion in APScheduler itself is the right choice. I might want to use a time series database to record job submissions from all job stores. Then again, not everybody might want to store such history for reasons of disk space or just data retention policies.

Changes of this scale would have to wait until v4.0 anyway, and my plan was to tinker with a new design using my other project ([asphalt-tasks](https://github.com/asphalt-framework/asphalt-tasks)) as a staging ground. I have beginnings of an alternate system of job submission tracking there, though history still requires event listeners saving them to some database. That scheduler is better designed with better decoupling and perhaps more suitable as a testing ground but is still quite incomplete due to me being swamped with work and other F/OSS projects.

seanharr11:
To your point about no one asking for job history retention in 7 years: that has never stopped any entity from building something they thought would benefit others, even if they weren't asking for it yet.

I also would suggest **not** use a different database (like a timeseries DB) to store job_submissions, as this would add another dependency to end-user's projects, especially when you do such a good job of supporting many different jobstores.

You bring up a valid concern of some end-users not wanting to persist job history due to disk space or data retention policies. I think the easy answer here to make this feature optional, that way, it is must less intrusive.

There is a lot of good work done here in this PR that I'd like to preserve, and I'd hate to see you re-create the wheel, especially if you had intentions to track job history in the future. Are you willing to work together, either in a **job-submission-feature-branch**, or in your **asphalt-tasks (v4.0) sandbox**, to bring this idea to fruition?

agronholm:
> To your point about no one asking for job history retention in 7 years: lack of a feature request has never stopped any team or individual from building new things.

> I also would suggest not using a different database (like a timeseries DB) to store job_submissions, as this would add another dependency to end-user's projects, especially when you do such a good job of supporting many different jobstores.

Which is a decision I've started to regret. I have less and less free time and the increasing amount of scheduler, executor and job store back ends significantly increases the maintenance burden, to the point where I simply don't have the energy to add any new features unless someone sends a polished PR, and even then with after a significant amount of scrutiny.

The code base is also starting to show its age. Python 2.7 compatibility is a big PITA but I don't want to drop it yet out of fear of alienating the user base. The test suite could use an overhaul, which I've begun in asphalt-tasks.

>You bring up a very valid concern of some end-users not wanting to persist job history due to disk space or data retention policies. I think the easy answer here to make this feature optional, that way, it is must less intrusive.

Amen to that. The reason why many people prefer APScheduler to Celery is its simplicity. This PR does not exactly advance that goal, though I see why some might want it. I prefer solutions like flask-apscheduler where someone else wraps it as a service to provide extra value, and this is partially why I added the events system in the first place.

Asphalt-tasks (I might name it asphalt-scheduler before the first release) adds a concept of a *ticket*. This seems equivalent to your concept of job submissions. Tickets can be awaited and cancelled. The events will also specify the ticket along with the expected run time(s) so they can be used to build the kind of service you're looking for. So I am +1 for tracking submissions but -1 for a built-in history feature.

My biggest problem with asphalt-tasks at the moment is how to handle distributed processing. I intend for this to become a viable alternative to Celery. My main concern here is the division of responsibilities between different Asphalt libraries. Should I keep asphalt-tasks purely as a scheduler library and implement task queueing elsewhere? This is the question I've been struggling with.

My focus has lately been on asphalt-web – my Asphalt based web application component. I really wanted to get that done before devoting any significant amount of time to anything else. Sadly, the last few weekends I've been mostly doing paid work. But that shouldn't stop you from investigating Asphalt and Asphalt-tasks. That is where I would welcome some help. If you're willing to work with me on that, I can hop on video chat and discuss this further.

seanharr11:
I would certainly like to hear more about Asphalt / Asphalt-tasks, and would be interested in collaborating with you on the project. The ticket system that you describe sounds very much like what I've implemented here. Where are you located? I have lots of time this Friday (tomorrow) to discuss over a video chat. I am in Boston, MA (EST).

Additionally, if I re-worked this PR such that it made this an optional parameter (i.e. ``track_job_submissions=True``), would you be willing to merge this? I know you want to avoid feature creep with APScheduler, but I would maintain this feature, and it wouldn't otherwise change any functionality of the tool. At the very least, we could get some feedback from folks about what they like/dislike about the feature, and this would very much help our design/implementation of the ticket system in Asphalt-Tasks.

agronholm:
I'm located in Finland (EET). I'm having people over on Friday but given the 6 hour time zone difference, maybe you can catch me afterwards?

Regardless of my eventual decision, a feature of this magnitude is going to have to be part of APScheduler 4.0, for which I had other job store system changes planned too, like getting rid of pickle in favor of other serialization schemes.

seanharr11:
That sounds very reasonable. We can discuss this too in the call.

And yeah that should work. What time (EET) will you be available? I believe I am 7 hours behind you, so it sounds like I can make anytime work if people leave your house at night.

agronholm:
11 – 11:30 pm would probably work. I might not be at my best but I'll try :)

seanharr11:
haha, nothing like talking Software after a couple of beers...

If you have an Apple device we could chat over Facetime, either that or Google Hangouts works for me. My email is seanharr11@gmail.com ... if you send me an email I'll have you on Hangouts, and I can send you my phone number if you want to use Facetime.

Does that work?

agronholm:
It's been a while, but I decided not to include this feature in the foreseeable future. Might be good as an add-on package though!

process0:
@agronholm I'm interested in implementing job history functionality.  It seems like it can be done using just events.

agronholm:
Most likely, yes.

PR #171, opened 2016-11-24
Update super call to support Python 2.7

orweis:
Quick fix in TornadoExecutor.__init__'s super() call to support Python 2.7


agronholm:
Thanks! I write super() by reflex these days I guess :/

PR #172, opened 2016-11-29
Trigger comparison

funkyfuture:
i needed to test the results of a parsing function that returns triggers. though i could have done this in other ways, i felt it would be intuitive to compare with expected instances.

however, one could argue that start-, endtime and timezone for an interval shouldn't be considered. as from a common point of view an hourly interval equals an hourly interval regardless the place on earth (don't know about marsian timezones though) an when it started. in my use-case that would apply.

`__lt__`  etc. could also be implemented for date and interval.

there's also a patch that allows to invoke `setup.py` directly, what i usually do and perceive as common.

### DIFF
--- a/apscheduler/triggers/base.py
+++ b/apscheduler/triggers/base.py
@@ -8,6 +8,17 @@ class BaseTrigger(six.with_metaclass(ABCMeta)):
 
     __slots__ = ()
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        for slot in self.__slots__:
+            if getattr(self, slot) != getattr(other, slot):
+                return False
+        return True
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
     @abstractmethod
     def get_next_fire_time(self, previous_fire_time, now):
         """
--- a/apscheduler/triggers/cron/fields.py
+++ b/apscheduler/triggers/cron/fields.py
@@ -70,6 +70,9 @@ class BaseField(object):
     def __eq__(self, other):
         return isinstance(self, self.__class__) and self.expressions == other.expressions
 
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
     def __str__(self):
         expr_strings = (str(e) for e in self.expressions)
         return ','.join(expr_strings)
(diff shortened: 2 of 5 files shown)


agronholm:
Merging this would add to my maintenance burden so I have to be very careful about what I add and what I don't. In all these years nobody else has asked for comparable triggers. Can you convince me that the use case is common enough to warrant this?

As for the setup.py thing, it is quite rare to see someone add hashbangs to setup.py. The vast majority of people don't care as they use pip to install libraries.

apscheduler/triggers/base.py
```diff
@@ -8,6 +8,17 @@ class BaseTrigger(six.with_metaclass(ABCMeta)):
 
     __slots__ = ()
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        for slot in self.__slots__:
```
agronholm: Doesn't this fail to account for other slots in the class hierarchy?

apscheduler/triggers/cron/fields.py
```diff
@@ -70,6 +70,9 @@ def compile_expression(self, expr):
     def __eq__(self, other):
         return isinstance(self, self.__class__) and self.expressions == other.expressions
 
+    def __ne__(self, other):
```
agronholm: Does this really have to be repeated in every subclass?

funkyfuture:
> In all these years nobody else has asked for comparable triggers.

i already wondered about it.

> Can you convince me that the use case is common enough to warrant this?

i'm not keen to convince you, especially not since the semantics aren't obvious for interval.

an argument would be that currently `DateTrigger('2016-12-29') == DateTrigger('2016-12-29')` returns `False` which seems odd.

apscheduler/triggers/base.py
agronholm: In the case of built-in triggers, sure. But what if someone subclasses those triggers and adds their own `__slots__`?

funkyfuture:
> it is quite rare to see someone add hashbangs to setup.py. 

i claim the opposite! :-D

agronholm:
I am likely to turn down this PR, as its usefulness is questionable and the semantics are unclear, as you said yourself.

funkyfuture:
no problem. as i said, there are other ways. i'll close it tomorrow if i have no further thought that may be of relevance.

oh, since i'm going to subclass triggers anyway, i could wrap the comparision capabilities in a mixin-class. that would also allow mixins with different semantics.

agronholm:
Sounds like a good idea.

funkyfuture:
i gave it a shot and concluded that it's less effort to deal with it in the client code than a general implementation.

PR #175, opened 2016-12-16
Changed type hint of :param seconds: to float

williamhatcher:
The scheduler can run at subsecond intervals, so I changed the type hint of :param seconds: in __init__ to float.


agronholm:
Sorry I took this long to look at this. But subsecond accuracy was never among the project goals. Also, float is not a superclass of int so technically this excludes ints from being valid arguments for `seconds` :)

agronholm:
What might be okay is `Union[int, float]` or [numbers.Real](https://docs.python.org/2.7/library/numbers.html?highlight=real#numbers.Real).

PR #191, opened 2017-03-26
Docs UserGuide Chinese  Translation

ppmoon:
I translated the docs of userguide to chinese maybe can help somebody.
I think maybe some translations isn't very accurate.But if I find wrong translation  I will correct it.


agronholm:
I'm not very comfortable including a translation I can neither check or update myself. Besides, I don't think everything inside docs/ should be duplicated for each translated language. Also, I don't see any Chinese text there, so where's the translation?

agronholm:
Perhaps an entry in the Wiki would suffice? I could just link to that.

ppmoon:
I just translate userguide.rst and index.rst. I think wiki link maybe good. 
Or When I finished all the documents I create a gitbook or read the doc book.Then I contact you ,you can link to the page.

agronholm:
I suggest you mention in the translation which version of APScheduler you used as source material. That way readers will know if the translated version is current or stale.

agronholm:
Let's not keep this PR open. Let me know when and if you've put the translation online so I can link to it.

PR #195, opened 2017-04-01
fix case where both qt4 and qt5 are installed

eendebakpt:
If both PyQt4 and PyQt5 are installed, but PyQt4 has already been imported the statement `from PyQt5.QtCore import QObject, QTimer` generates the following error:
```
RuntimeError: the PyQt5.QtCore and PyQt4.QtCore modules both wrap the QObject class
```
This PR fixes this by catching the RuntimeError generated and trying PyQt4.


agronholm:
Thanks!

eendebakpt:
@agronholm The PR was not correct I think. I made a small update in #196.

PR #196, opened 2017-04-02
use tuple in catching exceptions

eendebakpt:
This is a fix to #195.


agronholm:
I'm embarrassed that I didn't notice this myself.

PR #211, opened 2017-06-26
remove the module that have been imported before __import__

0x57:
After I add a job to the jobstore, I want to modify the job, so I remove the job. After modify my job code, and then re-add the job, but the job did not take effect, it still the implementation of the old code, I found that the __import__ problem, so I deleted the previously imported module before __import__ and then re-imported the modified module with __import__.  After this modification as long as I modify the module, the code will immediately take effect, and will not be the implementation of the old code.

### DIFF
--- a/apscheduler/util.py
+++ b/apscheduler/util.py
@@ -9,6 +9,8 @@ from functools import partial
 from pytz import timezone, utc
 import six
 
+import sys
+
 try:
     from inspect import signature
 except ImportError:  # pragma: nocover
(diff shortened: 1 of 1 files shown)


agronholm:
This change would be detrimental to performance. In the vast majority of cases, users *want* the module to be cached in sys.modules to avoid the overhead of re-importing. This change would likely have negative consequences as users currently do not expect their code to be reimported every time a job is executed.

agronholm:
Code reloading is generally a bad idea as it is extremely difficult to get right in any language. I ask that you find another way.

0x57:
Thank you for your reply, I agree with you, I will find another way. This is my first time to submit pull request, consider not comprehensive enough, next time I will pay more attention. Thanks again for your reply!

Issue #189, opened 2017-02-28
Scheduler getting stuck with some invalid cron rules


dribler:
Hi,

some cron rules causing scheduler run in infinite loop. Couple of example below

```
from apscheduler.schedulers.background import BackgroundScheduler
from apscheduler.triggers.cron import CronTrigger

scheduler = BackgroundScheduler()

scheduler.add_job(lambda: None, CronTrigger(day_of_week=8) )
# scheduler.add_job(lambda: None, CronTrigger(minute=60) )  # this would cause same
scheduler.start()
```
Older apscheduler versions treated values outside of the range as max/min possible value (which was perfectly fine for us). In our case that's a bit dangerous b/c it also doesn't raise any exceptions on invalid rules.

I believe this commit caused it https://github.com/agronholm/apscheduler/commit/4a23598815130a8a35022bcc4a6ffb4d9db04dd5, i would try to prepare PR to fix it, but feeling lack of time currently. 

I noticed that after
```
minval = max(minval, self.first)
```
`minval` equals 8 in case of `CronTrigger(day_of_week=8)` b/c `self.first == 8`. Hoping it helps.

PR #236, opened 2017-10-25
Cron expression will now raise ValueError when expression values are …

ofirule:
…out of range (#189)


ofirule:
Hi @agronholm please have a look, I think this is a nice fix/feature

agronholm:
Yes, I will take a closer look during the weekend.

agronholm:
While you wait, please fix the long lines (>99 cols) in your PR.

agronholm:
While I agree that this should be done, I'm not sure this is the way after looking at the implementation. I was almost sure I had something before for checking that the value is within the limits. At any rate, the tests should be done using py.test's parametrization feature. But there's no rush, I might even create my own patch instead of merging this one.

agronholm:
I ended up fixing this my own way. Thanks for your effort though.

ofirule:
Great, I will definitely use it

PR #257, opened 2017-12-08
Fix memory leak when job raises an exception

### LINKED ISSUE #235: Memory leak when worker raises an exception
Hello,

If a worker raises an exception and it isn't handled, the memory used by the worker will not be released.

This can be reliably reproduced by the following code snippet:

```python
from apscheduler.schedulers.background import BlockingScheduler

def bar():
    a = [{i: i + 2} for i in range(10000000)]
    for (k, v) in a.items():
        a[k] = v + 5

def foo():
    scheduler = BlockingScheduler({
        'apscheduler.executors.default': {
            'class': 'apscheduler.executors.pool:ThreadPoolExecutor',
            'max_workers': '20'
        },
        'apscheduler.job_defaults.coalesce': 'false',
        'apscheduler.job_defaults.max_instances': '1',
        'apscheduler.timezone': 'UTC'
    })

    scheduler.add_job(bar, trigger='cron', minute='*')
    scheduler.start()

foo()
```

Thanks

gilbsgilbs:
Fixes #235.

It cleans traceback frames where required to avoid keeping cyclic references everywhere thus preventing GC from happening.

If you're concerned by the `__traceback__` cleanup for py3 (for performance reasons or because it might be a tiny tiny breaking change), I can implement the `__del__` way, but it's much more intrusive. Unless you have a better way around this.

### DIFF
--- a/apscheduler/executors/base.py
+++ b/apscheduler/executors/base.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta
 from traceback import format_tb
 import logging
 import sys
+import traceback
 
 from pytz import utc
 import six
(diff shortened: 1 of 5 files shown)


gilbsgilbs:
Tests indicates that GC doesn't work with py2 and pypy. This needs a bit more work appearently. Some help is welcome :) .

agronholm:
A word of warning, I have a half finished patch which is *vastly* simpler than this one.

gilbsgilbs:
@agronholm Thanks for letting me know. Let's close this for now then.

PR #258, opened 2017-12-10
Implement random jitter option for CronTrigger and IntervalTrigger

### LINKED ISSUE #168: Feature request: allow interval jobs to be offset by a random value
When jobs are configured to execute at certain times, they tend to coalesce at one particular time. For example, executing 10 jobs at the top of the hour, another 10 once in at noon and once at midnight, and a third job that executes every minute will cause all of these jobs to run at the same time at noon and midnight, potentially causing a bottleneck in the number of threads.

This could be solved by introducing a new parameter to the interval trigger that would define a value in seconds by which each job would be offset.

Assuming these new imports:
``` python
from operator import add, sub
from random import randrange, choice
```
The constructor could have a new parameter value (`random_offset`) that would be at most equivalent to the configured interval:
``` python
self.random_offset = min(self.interval_length, random_offset)
```
Then, the `get_next_fire_time()` function could check if an offset is defined and modify the next execution time accordingly:
``` python
if self.random_offset:
    offset = timedelta(seconds=randrange(0, self.random_offset))
    op = choice((add, sub))
    next_fire_time = max(datetime.now(self.timezone), op(next_fire_time, offset))
```
I understand all this can be worked around by setting custom start times for each group of jobs so that they don't intersect but having an easy-to-configure parameter would be cool too. The UNIX Anacron scheduler uses a similar mechanic with the [`RANDOM_DELAY` option](http://man7.org/linux/man-pages/man5/anacrontab.5.html).

I did notice during testing of the above code that `get_next_fire_time()` is called both from `_get_run_times()` (in `job.py`) and `_process_jobs` in (in `base.py`), producing different times for each. I'm not sure if it's significant that these functions both receive the same next fire time.

Anyway, is this something you'd consider merging if I were to submit a PR? Thank you!

gilbsgilbs:
Fixes #168

### DIFF
--- a/apscheduler/triggers/base.py
+++ b/apscheduler/triggers/base.py
@@ -1,5 +1,5 @@
 from abc import ABCMeta, abstractmethod
-import datetime
+from datetime import timedelta
 import random
 
 import six
@@ -37,7 +37,7 @@ class BaseTrigger(six.with_metaclass(ABCMeta)):
         if next_fire_time is None or not jitter:
             return next_fire_time
 
-        next_fire_time_with_jitter = next_fire_time + datetime.timedelta(
+        next_fire_time_with_jitter = next_fire_time + timedelta(
                 seconds=random.uniform(-jitter, jitter))
 
         if next_fire_time_with_jitter < now:
--- a/apscheduler/triggers/base.py
+++ b/apscheduler/triggers/base.py
@@ -23,7 +23,7 @@ class BaseTrigger(six.with_metaclass(ABCMeta)):
     def _apply_jitter(self, next_fire_time, jitter, now):
         """
         Randomize ``next_fire_time`` by adding or subtracting a random value (the jitter). If the
-        resulting datetime is in the past, returns ``now``.
+        resulting datetime is in the past, returns the initial ``next_fire_time`` without jitter.
 
         ``next_fire_time - jitter <= result <= next_fire_time + jitter``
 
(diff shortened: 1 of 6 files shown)


apscheduler/triggers/base.py
```diff
@@ -17,3 +19,30 @@ def get_next_fire_time(self, previous_fire_time, now):
...
+            ``None``, returns ``None``.
+        :param int|None jitter: maximum number of seconds to add or subtract to
+            ``next_fire_time``. If ``None`` or ``0``, returns ``next_fire_time``
+        :param datetime.datetime now: current datetime
+        :return datetime.datetime|None: next fire time with a jitter.
+        """
+        if next_fire_time is None or not jitter:
+            return next_fire_time
+
+        next_fire_time_with_jitter = next_fire_time + datetime.timedelta(
```
agronholm: My standard has been to import the classes (like `timedelta`) directly.

apscheduler/triggers/interval.py
```diff
@@ -20,12 +20,14 @@ class IntervalTrigger(BaseTrigger):
     :param datetime|str start_date: starting point for the interval calculation
     :param datetime|str end_date: latest possible date/time to trigger on
     :param datetime.tzinfo|str timezone: time zone to use for the date/time calculations
+    :param int|None jitter: randomize the job execution time. Job will be advanced or delayed by
```
agronholm: Should probably be rephrased as "advance or delay the job execution by `jitter` seconds"

apscheduler/triggers/interval.py
```diff
@@ -83,10 +91,12 @@ def __setstate__(self, state):
         self.end_date = state['end_date']
         self.interval = state['interval']
         self.interval_length = timedelta_seconds(self.interval)
+        self.jitter = state.get('jitter')
```
agronholm: Check the version number here too.

agronholm:
The changes I requested for the interval trigger of course apply to the cron trigger as well.

apscheduler/triggers/interval.py
```diff
@@ -65,7 +72,8 @@ def __getstate__(self):
             'timezone': self.timezone,
             'start_date': self.start_date,
             'end_date': self.end_date,
-            'interval': self.interval
+            'interval': self.interval,
+            'jitter': self.jitter,
```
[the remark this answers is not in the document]
gilbsgilbs: Yup. Didn't know what to do since it's not a breaking change. Should be fine now.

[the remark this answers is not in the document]
agronholm: The version number is incremented because jobs that have jitter will no longer work as intended on older versions of APScheduler.

agronholm:
Thank you!

PR #260, opened 2017-12-13
Allow multiple spaces in crontab format

luc-tielen:
I noticed in commit 39cb1a5f65a2d5970b9639e3e324ad7953d261f9 that expr.split(' ') was used i.s.o. expr.split().
If .split() is used; then it is also possible for multiple spaces to be added in between (or in front).

Simple example showcasing different behavior:

```python
>>> 'a  b'.split()
['a', 'b']
>>> 'a  b'.split(' ')
['a', '', 'b']
>>> ' a  b'.split(' ')
['', 'a', '', 'b']
>>> ' a  b'.split()
['a', 'b']
```

This pull request changes the code to use .split().

### DIFF
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -95,7 +95,7 @@ class CronTrigger(BaseTrigger):
         :return: a :class:`~CronTrigger` instance
 
         """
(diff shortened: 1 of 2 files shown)


agronholm:
Thanks.

PR #268, opened 2018-01-02
Prevent deadlocks in scheduler shutdown

pR0Ps:
By splitting up the executor lock and jobstore lock, this allows jobs
that are shutting down to query their state from the database (requiring
`self._jobstore_lock`) without causing a deadlock.

Since setting `self.state` above prevents new jobs from being added to the
executors as they shut down, breaking up these locks won't allow new
jobs to run while the scheduler is shutting down.

### DIFF
--- a/apscheduler/schedulers/base.py
+++ b/apscheduler/schedulers/base.py
@@ -177,12 +177,13 @@ class BaseScheduler(six.with_metaclass(ABCMeta)):
 
         self.state = STATE_STOPPED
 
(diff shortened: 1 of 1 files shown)


agronholm:
Do you think you could add a test to validate this fix?

agronholm:
I figured the test would be pretty challenging to add, and the code change seems simple enough that I could merge it without too many doubts. Thank you!

PR #270, opened 2018-01-03
Fix CronTrigger sometimes fired after end_date when jitter is enabled.

### LINKED ISSUE #269: CronTrigger with jitter may execute after end_date
To reproduce, next fire time without jitter must be before `end_date`, and the jitter make the next fire time exceed `end_date`. `IntervalTrigger` does not seem to be affected by this.

gilbsgilbs:
Fixes #269

### DIFF
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -194,6 +194,10 @@ class CronTrigger(BaseTrigger):
         if fieldnum >= 0:
             if self.jitter is not None:
                 next_date = self._apply_jitter(next_date, self.jitter, now)
+
+                if self.end_date and next_date > self.end_date:
+                    return None
+
             return next_date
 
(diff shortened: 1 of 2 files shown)


apscheduler/triggers/cron/__init__.py
```diff
@@ -194,6 +194,10 @@ def get_next_fire_time(self, previous_fire_time, now):
         if fieldnum >= 0:
             if self.jitter is not None:
                 next_date = self._apply_jitter(next_date, self.jitter, now)
+
+                if self.end_date and next_date > self.end_date:
+                    return None
```
agronholm: No. Applying jitter should not discard an otherwise fine fire time. If you want to fix this, limit the datetime by end_date after applying jitter but don't discard it.

agronholm:
I've incorporated your tests to my own fix and credited you accordingly.

PR #272, opened 2018-01-03
Fix test failing randomly

gilbsgilbs:
This test happened to be failing for very low jitter values such as `-2.999…`. This is because `IntervalTrigger` takes its own `now` value as `start_date`, which is slightly different from the `now` value inside the test. Consequently, the next fire time calculated could have been in less than two seconds we didn't consider the same base value in the test as the trigger.


agronholm:
I think given #416, this is no longer relevant so I'm closing the PR. Let me know if I was in error.

PR #277, opened 2018-01-18
Bump fallback version

mcmclx:
Looks like this was missed - also shows up in the releases in pypi.


agronholm:
Where on PyPI does the fallback version show up? There is no point in updating this with every release, as otherwise I would just manually write the version every time instead of relying on git tags.

agronholm:
This feels like a mistake so I'm closing this. If you feel you can offer a better way to provide a fallback release version (without always updating it in two places), then by all means reopen or send another PR.

PR #316, opened 2018-06-01
gevent is already available for Python 2.x and 3.x

outime:
I noticed this outdated tip in the docs - gevent already supports Python 3:

>gevent 1.3 runs on Python 2.7 and Python 3. Releases 3.4, 3.5 and 3.6 of Python 3 are supported.


agronholm:
Good catch. The test suite has already embraced gevent on Python 3 a long time ago.

PR #317, opened 2018-06-07
fix bug: 'undefined' is not None

chmberl:


agronholm:
Wait, no value there is ever supposed to be `undefined`. In what circumstances would this happen?

chmberl:
https://github.com/agronholm/apscheduler/blob/master/apscheduler/schedulers/base.py#L363-L366
the misfire_grace_time default value is `apscheduler.util.undefined`, so 'job.misfire_grace_time is not None' will be True, https://github.com/agronholm/apscheduler/blob/master/apscheduler/executors/base.py#L114.
If there are multiple jobs executing at the same time，some jobs will be missed if the misfire_grace_time is not set.

agronholm:
No, there is a default value for `misfire_grace_time` (1) which is set if the value during `add_job()` was not passed. Are you sure that job.misfire_grace_time is actually `undefined`?

chmberl:
> If there are multiple jobs executing at the same time，some jobs will be missed if the misfire_grace_time is not set

I mean that misfire_grace_time set to a default value. the default value is apscheduler.util.undefined
```
 if job.misfire_grace_time is not None:
        difference = datetime.now(utc) - run_time
        grace_time = timedelta(seconds=job.misfire_grace_time)

#   job.misfire_grace_time(apscheduler.util.undefined) is not None ?
```

agronholm:
The `misfire_grace_time` attribute in a Job can never be `undefined`. The only place where `undefined` is used is in the scheduler to indicate that the default value should be used which, in the case of `misfire_grace_time` is 1, as I already pointed out.

agronholm:
I'm closing this until you can prove that this can actually happen somehow.

chmberl:
> The misfire_grace_time attribute in a Job can never be undefined

the default value is apscheduler.util.undefined.

chmberl:
image

agronholm:
What do you suppose this screenshot proves? All it tells me is that the job was missed by more than 1 second. If this was really a problem, a lot of people would be reporting it. Have you read my previous messages? I explained that apscheduler uses the default misfire grace time from it settings which is 1 by default. Why do I have to repeat myself?

PR #318, opened 2018-06-12
make it compatible for python 2.6

garyelephant:
This line will raise SyntaxError in Python2.6.6

```
values = {k: int(v or 0) for k, v in values.items()}
```


agronholm:
Python 2.6 has reached its End Of Life in 2013. I will not start supporting dead Python versions.

PR #323, opened 2018-07-25
Add milliseconds support for the interval scheduler

aradler:
Because of the way you implemented this (using timedeltas), the change involves very few lines of code.

I opted out of implementing this as microseconds, as it is unlikely that a scheduler running on any normal computer could actually uphold microsecond precision.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ docs/_build/
 build/
 virtualenv/
 example.sqlite
+.pytest_cache
--- a/apscheduler/triggers/interval.py
+++ b/apscheduler/triggers/interval.py
@@ -17,6 +17,7 @@ class IntervalTrigger(BaseTrigger):
     :param int hours: number of hours to wait
     :param int minutes: number of minutes to wait
     :param int seconds: number of seconds to wait
+    :param int milliseconds: number of milliseconds to wait
     :param datetime|str start_date: starting point for the interval calculation
     :param datetime|str end_date: latest possible date/time to trigger on
     :param datetime.tzinfo|str timezone: time zone to use for the date/time calculations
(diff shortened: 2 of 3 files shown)


agronholm:
The same would hold true for milliseconds too, at least when using a persistent job store. What is your use case for millisecond accuracy anyway?

aradler:
I am interfacing with an API that allows three requests per second, and I would like to be able to request as much data as possible.

I installed my fork's version of the package into my project, and found that I have roughly +/- 2 milliseconds on most queries. I haven't done any long haul or heavy load testing; are there any tests that I could add that you think would provide a useful insight to the precision in all cases? I'm considering writing a test case that simply starts a few timers with some millisecond duration, lets them all run and log the times that they actually run at, and ensuring that the precision is within some factor. My biggest concern here, though, is that test would likely run different on different (especially slower) platforms.

agronholm:
I will consider this for the next release.

gusutabopb:
+1

I was looking for this exact functionality so it'd be really nice to have it in a proper release.

My use case is that need to run some jobs with a 100~500ms frequency for a trading system.

`DateTrigger` already supports sub-second triggers (since it takes any datetime object), this PR should add `IntervalTrigger` support and adding `CronTrigger` support also doesn't seem too hard (I can submit another PR for that if the library author thinks this would be nice to have).

agronholm:
I'm hoping to set some time aside this summer to work on APScheduler 4.0. Note that the scheduler is not real-time and won't be in the foreseeable future, but the ability to specify execution times on the millisecond level has been requested by quite a few people so I think I'll just go ahead and do it. Just don't complain if the jobs don't exactly start at the designated sub-second times.

thedrow:
We need this for Celery 5.

agronholm:
Millisecond intervals are supported in master.

PR #329, opened 2018-08-10
Fix jobs with duplicate id when scheduler is not ran

mohi7solanki:
ref: https://stackoverflow.com/questions/51787976/apscheduler-job-ids-why-are-jobs-allowed-to-share-ids/51791145

Duplicate jobs with the same id can be created if the scheduler has not start, we can check early to prevent it.

### DIFF
--- a/apscheduler/schedulers/base.py
+++ b/apscheduler/schedulers/base.py
@@ -409,6 +409,14 @@ class BaseScheduler(six.with_metaclass(ABCMeta)):
         :rtype: Job
 
         """
+
+        # Check if job with `id` already exists if `id` is provided.
+        if id is not None and not replace_existing:
+            # `get_job` should return None if there is no job with this id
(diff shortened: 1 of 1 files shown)


agronholm:
But the job stores haven't even been started at that point.

mohi7solanki:
@agronholm then is there any other way we could give a warning about jobs with duplicate id?

mohi7solanki:
Okay, can we show warnings about this?

agronholm:
You would rather show warnings than actually provide a solution? And based on what should we show warnings?

mohi7solanki:
I have provided the solution.. I am raising ConflictingIdError if job with same id exists in when scheduler is not ran

agronholm:
Such a solution is only limited to pending jobs though. You will still have problems if the scheduler is started and a conflict occurs in a persistent job store.

PR #335, opened 2018-10-01
peewee jobstore

njason:


agronholm:
While the code looks good at a glance, I'm not willing to accept it as part of APScheduler itself because the SQLAlchemy job store is already sufficient for providing RDBMS support. Therefore I don't really see the value in a Peewee job store. Feel free to publish this as an add-on though!

njason:
https://gist.github.com/njason/df77960dc06f9745bb9421c6b9e6b9f0

PR #345, opened 2018-11-27
Docs: add `py-modindex` to side bar (fix #343)

### LINKED ISSUE #343: docs: add api reference to left navigation pane
[API reference](https://apscheduler.readthedocs.io/en/latest/py-modindex.html) is most useful day-to-day  page (after you pass tutorial). Second time I spent 5 minutes to find link to this page, since it is not on left navigation pane. Sugget to add it.

pohmelie:
fix #343


agronholm:
Did something appear on the left hand panel for you? Didn't for me.

agronholm:
Actually it seems that using `sphinx_rtd_theme` does make the TOC appear on the side bar. I should probably use that even for local builds by default.

pohmelie:
If this theme is pip instalable, then we can try it localy.

pohmelie:
Yeah, looks like it works without full toc extension. I will fix that.

.readthedocs.yml
```diff
@@ -11,3 +11,4 @@ python:
     - tornado
     - twisted
     - zookeeper
+    - docs
```
agronholm: Rename this to "doc" (I like to stay consistent with my other projects).

docs/index.rst
```diff
@@ -17,6 +17,7 @@ Table of Contents
   contributing
   extending
   faq
+  py-modindex
 
 
 Indices and tables
```
agronholm: You can remove the second API reference link now.

setup.py
```diff
@@ -61,7 +61,11 @@
         ],
         'testing:python_version == "2.7"': ['mock'],
         'testing:python_version == "3.4"': ['pytest_asyncio < 0.6'],
-        'testing:python_version >= "3.5"': ['pytest_asyncio']
+        'testing:python_version >= "3.5"': ['pytest_asyncio'],
+        'docs': [
```
agronholm: Rename this to "doc".

agronholm:
Thanks! This has been somewhat of a PITA at times.

PR #347, opened 2018-12-04
Import Iterable, Mapping from collections.abc

romanlevin:
```
.../lib/python3.7/site-packages/apscheduler/job.py:1
  .../lib/python3.7/site-packages/apscheduler/job.py:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Iterable, Mapping
```

### DIFF
--- a/apscheduler/job.py
+++ b/apscheduler/job.py
@@ -1,9 +1,3 @@
-try:
-    # Importing ABCs from collections instead of collections.abc
-    # will stop working in Python 3.8.
-    from collections.abc import Iterable, Mapping
-except ImportError:
-    from collections import Iterable, Mapping
 from inspect import ismethod, isclass
 from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


agronholm:
Why a conditional import?

agronholm:
Oh, right, Python 2.7 doesn't have the `collections.abc` module.

romanlevin:
> Oh, right, Python 2.7 doesn't have the `collections.abc` module.

I actually didn't check, but I assumed it would fail on at least some older supported versions.

agronholm:
Drop the comment and move the block to after the other imports (PEP 8 requires that conditional imports are added last). I'll merge then

romanlevin:
@agronholm Done – also spotted another ABC import in `apscheduler/schedulers/base.py`.

agronholm:
Thanks!

romanlevin:
@agronholm btw, is there ongoing work anywhere to fix the failing tests?

agronholm:
I've been meaning to do that my lately my free time has been quite constrained. This is a high priority issue for me though so I expect to fix it this week. From the looks of it, the redis library seems to have changed its API.

agronholm:
I've fixed the build. The culprit was a new major release of the redis library which had broken backwards compatibility.

PR #350, opened 2018-12-13
Correct update_job  to raise with job id

Ugrend:
I believe this should be raising with job.id not the builtin id


agronholm:
Good catch, thanks!

PR #357, opened 2019-01-30
Update FAQ to Explain Common ValueErrors

JeffVandrewJr:
One of the most common questions on StackOverflow (and the issues page here) revolves around trying to store jobs in a persistent job store and getting ValueErrors. This addition hopes to head off those questions.


agronholm:
This is missing the fact that class and static methods are scheduleable, at least on Python 3.

JeffVandrewJr:
@agronholm I just added that extra info at your suggestion.

JeffVandrewJr:
@agronholm My original edit had the classmethod info backwards; I just updated it so we should be good now.

PR #363, opened 2019-03-06
Fix signature inspection for wrapped function

meownoid:
Wrapped functions can not be scheduled because of signature inspection. This little fix makes it possible to schedule wrapped functions.

Simple example:
```python
from functools import wraps

from apscheduler.executors.pool import ProcessPoolExecutor
from apscheduler.jobstores.memory import MemoryJobStore
from apscheduler.schedulers.background import BlockingScheduler


def wrapper(fn):

    @wraps(fn)
    def f(prefix, x):
        return prefix + str(fn(x))

    return f


@wrapper
def f(x):
    return 2 * x + 5


scheduler = BlockingScheduler(
    jobstores={
        'default': MemoryJobStore()
    },
    executors={
        'default': ProcessPoolExecutor(
            max_workers=1
        )
    },
    job_defaults={
        'coalesce': False,
        'max_instances': 1
    },
    timezone='UTC'
)

scheduler.add_job(
    'test:f',
    kwargs={
        'prefix': 'number_',
        'x': 2
    }
)
scheduler.start()
```

This code fails with the error:
```
ValueError: The target callable does not accept the following keyword arguments: prefix
```

But it works with `follow_wrapped=False`.

### DIFF
--- a/apscheduler/util.py
+++ b/apscheduler/util.py
@@ -332,7 +332,7 @@ def check_callable_args(func, args, kwargs):
     has_varargs = has_var_kwargs = False
 
     try:
-        sig = signature(func)
+        sig = signature(func, follow_wrapped=False)
     except ValueError:
         # signature() doesn't work against every kind of callable
         return


agronholm:
Can you rebase this against 3.x? Thanks.

agronholm:
I've made these changes in 5ef5aed and credited you in the changelog. Thanks.

PR #364, opened 2019-03-11
add actual_start_time to JobExecution events

robinfrick:
Heya, this is a great project. I especially love the events for logging. For my use case I would love to have the actual start time of a job exposed, additional to the scheduled run time. The use case for me would be to calculate how far my queue is lagging behind, how long an individual job took, and possibly more. 

If you think this addition makes sense and you are happy with my PR I'd be happy if this gets added. If not I'd be happy to hear some feedback nevertheless.

### DIFF
--- a/apscheduler/events.py
+++ b/apscheduler/events.py
@@ -85,10 +85,20 @@ class JobExecutionEvent(JobEvent):
     :ivar traceback: a formatted traceback for the exception
     """
 
-    def __init__(self, code, job_id, jobstore, scheduled_run_time, retval=None, exception=None,
-                 traceback=None):
+    def __init__(
+        self,
+        code,
+        job_id,
+        jobstore,
+        scheduled_run_time,
+        actual_start_time,
+        retval=None,
+        exception=None,
+        traceback=None,
+    ):
         super(JobExecutionEvent, self).__init__(code, job_id, jobstore)
         self.scheduled_run_time = scheduled_run_time
         self.retval = retval
         self.exception = exception
         self.traceback = traceback
+        self.actual_start_time = actual_start_time
--- a/apscheduler/events.py
+++ b/apscheduler/events.py
@@ -85,17 +85,8 @@ class JobExecutionEvent(JobEvent):
     :ivar traceback: a formatted traceback for the exception
     """
 
-    def __init__(
-        self,
-        code,
-        job_id,
-        jobstore,
-        scheduled_run_time,
-        actual_start_time,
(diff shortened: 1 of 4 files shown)


agronholm:
I'm open to this idea, but do not change my code style while adding to it. It makes it harder to spot the changes. In fact do not change my code style at all.

agronholm:
Also, two calls to `datetime.now()` are not necessary so you can reuse `actual_start_time` there.

robinfrick:
Hey @agronholm thanks for getting back to me and congrats here on this really great package, I just started using it, and I love it. 

Re change of style: Sorry my linter  got in the way. I'll revert those formatting changes. 

Re your comment about triggering datetime.now() twice, just to clarify, do you mean that in 
`difference = datetime.now(utc) - run_time` I should reuse it, or where?

agronholm:
> Re your comment about triggering datetime.now() twice, just to clarify, do you mean that in
difference = datetime.now(utc) - run_time I should reuse it, or where?

Yes, this is what I meant.

yj97595077:
Hi @robinfrick I tested your code in this PR, some comments:
1. Why not add the `actual_end_time` in this PR? The end time of the job is very useful in many case.
2. The `actual_start_time` should consider the time zone, otherwise it is a little strange with `run_time`.
Finally, thanks your idea about adding the actual time for job events.

agronholm:
This problem has been solved differently in master. The context-local job info now contains a property that gives you the amount of calculated jitter and the original scheduled time.

agronholm:
Ah, right, this was about the execution events. I will keep this PR open until I've made sure that this information is present in those too.

oleg-gerasimovich:
Hey! Thank you for these project, really great idea and implementation. I just have one question. In my project I'm using it in async function to start function when client triggers endpoint, and as this function runs quite long, I'm sending two different messages: `job is still in progress`, 201 code or `job finished`, 200 code. In this solution I'm trying to reach scheduler job status, like 'started, running, failed, succeed etc.', but in scheduler.events and in Job events currently implemented events can not give me such responses.

Regarding this I have a question: is there any other way to get job status to check that it is still running?

agronholm:
There are now fields named `scheduled_start` in `JobAcquired` and `JobReleased`, and a `started_at` field in `JobReleased` (8feb98fa0e671f85049603694663b7bb471be974). This should hopefully cover this use case in v4.x. I won't add any new features to the 3.x series at this point.

Issue #366, opened 2019-03-21
the method inherited from base class can not be executed by processpool executor


yj97595077:
## Environment
My OS is macOS Mojave(10.14.3). The python and apscheduler is as blew.
```
# python -V
Python 2.7.10
```
```
# pip freeze
APScheduler==3.6.0
funcsigs==1.0.2
futures==3.2.0
pytz==2018.9
six==1.12.0
tzlocal==1.5.1
```

## Steps to Reproduce
I write one simple example to reproduce the problem.
`sched.py` is python main file, `base_action.py` and `action.py` are apscheduler job action.

- sched.py

```
# -*- coding: utf-8 -*-
import logging
import os
from apscheduler.schedulers.blocking import BlockingScheduler
from action import Action1, Action2

if __name__ == '__main__':
    logging.basicConfig()
    scheduler = BlockingScheduler()
    scheduler.add_executor('processpool')
    scheduler.add_job(Action1.run, 'interval', seconds=3)
    scheduler.add_job(Action2.run, 'interval', seconds=3)
    print('Press Ctrl+{0} to exit'.format('Break' if os.name == 'nt' else 'C'))

    try:
        scheduler.start()
    except (KeyboardInterrupt, SystemExit):
        pass
```

- base_action.py
```
# -*- coding: utf-8 -*-
from datetime import datetime

class BaseAction(object):
    @classmethod
    def run(cls):
        print('[%s]: The time is [%s]' % (cls.__name__, datetime.now()))
```

- action.py
```
# -*- coding: utf-8 -*-
from base_action import BaseAction

class Action1(BaseAction):
    pass

class Action2(BaseAction):
    pass
```

## Expected Behavior
I expect I can get the result as blew.
```
# python /Users/yangjian/Code/python-test/sched.py
Press Ctrl+C to exit
[Action1]: The time is [2019-03-21 12:35:45.060848]
[Action2]: The time is [2019-03-21 12:35:45.061480]
[Action1]: The time is [2019-03-21 12:35:48.060196]
[Action2]: The time is [2019-03-21 12:35:48.060664]
[Action1]: The time is [2019-03-21 12:35:51.060686]
[Action2]: The time is [2019-03-21 12:35:51.061235]
[Action1]: The time is [2019-03-21 12:35:54.060251]
[Action2]: The time is [2019-03-21 12:35:54.061208]

Process finished with exit code 0
```

## Current Behavior
Now, when I execute the program, I get the exception as blew.
```
[...]

PR #367, opened 2019-03-21
bugfix: the method inherited from base class can not be executed by p…

yj97595077:
Issue: https://github.com/agronholm/apscheduler/issues/366

### DIFF
--- a/apscheduler/util.py
+++ b/apscheduler/util.py
@@ -5,7 +5,7 @@ from __future__ import division
 from datetime import date, datetime, time, timedelta, tzinfo
 from calendar import timegm
 from functools import partial
-from inspect import isclass
+from inspect import isclass, ismethod
 import re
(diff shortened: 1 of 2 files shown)


agronholm:
Please add a test that would fail without this fix.

yj97595077:
> Please add a test that would fail without this fix.

OK, I try to add the testcase.

yj97595077:
> You know that you can use `tox` to run the tests locally, right? You don't have to rely on Travis for that.

Sorry about that. I will use `tox` to run the tests on my computer before pushing the commits from now on.

agronholm:
Thanks!

agronholm:
@yj97595077 I have to ask: why did you make the test so it is skipped on Python 3.2 and above? The reason does not make any sense either.

Issue #297, opened 2018-03-20
Feature request: Passing the scheduled start time to jobs


AdrianTeng:
Currently there are no way for the job to know what is the scheduled start time (only the current time by doing `datetime.now()`). This feature would be useful for time sensitive jobs (e.g. extra handling if the job is delayed)

DeeVeX:
http://apscheduler.readthedocs.io/en/latest/modules/triggers/cron.html#module-apscheduler.triggers.cron
Isn't "start_date" what you are looking for?

AdrianTeng:
@DeeVeX Nope. From what I understand from the docs, `start_date` is for setting the `CronTrigger`'s first trigger time. Whereas I want to know for each job, what is the scheduled trigger time, and passing that into the job

DeeVeX:
Oh that would be nice.
I could use the same thing for a jobs last run to do cleanups!

zhipcui:
Does anyone working on this?

viiicky:
Yes, this feature is useful to have(I need this in my current project). Is there any work going on this? I am also open to submit PR for this if I get some light on which direction to move on.

Till then a workaround that I can think of is:
1. While adding job(`add_job()`) to scheduler pass an explicit identifier and store it somewhere.
2. When the job runs, fetch job(`get_job()`) from scheduler making use of the stored id and then do:
job.next_run_time - trigger's value
Something of that sort should work.

But I am not sure if it is full proof. And also, what would happen in the scenarios when using DB store and the scheduler goes down and jobs are missed and then the scheduler comes up. Would the calculation be still correct in that case

Either way waiting for this feature...

viiicky:
Also I guess calling it `scheduled_time` instead of `scheduled_start_time` would be less confusing.
@AdrianTeng  @DeeVeX

viiicky:
Mine use case by the way is explained below:

Suppose a job is scheduled to poll data from a server every 10 minutes.
Now the request that is made to server might have a `start_time` and `end_time` to be mentioned in the request being made. Something like: `fetch_messages_from_smtp_server(start_time, end_time)` or an equivalent REST API maybe.

So, I will `add_job` to scheduler mentioning the start time say 7th Sept 2018 12:00:00 UTC and an interval of 10 minutes.
And inside the job definition currently I would do something on these lines (psuedocode):
```
start_time = datetime.utcnow()
end_time = start_time + 10 mins
fetch_messages_from_smtp_server(start_time, end_time)
```
[...]

PR #380, opened 2019-07-01
Implement issue #297 - Passing the scheduled start time to jobs

anon-john:
This PR implements the feature request from #297. Proposed changes were tested against py36, py37, py27 and flake8 with memjobstore and sqlalchemyjobstore stores. At the moment, unfortunately, I'm not able to test with mongo/redis/zookeeper/rethinkdb.

### DIFF
--- a/apscheduler/executors/base.py
+++ b/apscheduler/executors/base.py
@@ -120,6 +120,9 @@ def run_job(job, jobstore_alias, run_times, logger_name):
                 logger.warning('Run time of job "%s" was missed by %s', job, difference)
                 continue
 
+        if job.provide_scheduled_run_time:
+            job.kwargs["scheduled_run_time"] = run_time
+
         logger.info('Running job "%s" (scheduled at %s)', job, run_time)
         try:
             retval = job.func(*job.args, **job.kwargs)
--- a/apscheduler/job.py
+++ b/apscheduler/job.py
@@ -40,7 +40,7 @@ class Job(object):
 
     __slots__ = ('_scheduler', '_jobstore_alias', 'id', 'trigger', 'executor', 'func', 'func_ref',
                  'args', 'kwargs', 'name', 'misfire_grace_time', 'coalesce', 'max_instances',
-                 'next_run_time')
+                 'next_run_time', 'provide_scheduled_run_time')
 
     def __init__(self, scheduler, id=None, **kwargs):
         super(Job, self).__init__()
@@ -155,6 +155,10 @@ class Job(object):
         if 'func' in changes or 'args' in changes or 'kwargs' in changes:
             func = changes.pop('func') if 'func' in changes else self.func
             args = changes.pop('args') if 'args' in changes else self.args
+            provide_scheduled_run_time = changes.pop('provide_scheduled_run_time') \
+                if 'provide_scheduled_run_time' in changes \
+                else self.provide_scheduled_run_time
+
             kwargs = changes.pop('kwargs') if 'kwargs' in changes else self.kwargs
 
             if isinstance(func, six.string_types):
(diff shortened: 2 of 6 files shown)


anon-john:
It looks, that a single test (test_job_provide_scheduled_run_time) fails in travis. Surprisingly, running locally is OK. Any ideas/suggestions?

anon-john:
I executed the following to verify, that there are no missing changes in the branch of this PR:
```
git clone https://github.com/anon-john/apscheduler.git
cd apscheduler/
git checkout provide_scheduled_run_time
docker-compose up

# in a new session
python3.6 -m venv venv
source venv/bin/activate
pip install pytest
cd apscheduler/ # i.e. run against the 'provide_scheduled_run_time' branch
tox
```
With the above method, all tests run fine (i.e. there are no failures). The venv-related steps  are meant to provide modern pytest. Output of pip freeze is the following:
```
atomicwrites==1.3.0
attrs==19.1.0
importlib-metadata==0.18
more-itertools==7.1.0
packaging==19.0
pkg-resources==0.0.0
pluggy==0.12.0
py==1.8.0
pyparsing==2.4.0
pytest==5.0.0
six==1.12.0
wcwidth==0.1.7
zipp==0.5.1
```

agronholm:
I've hatched a more ambitious plan to implement this in APScheduler 4.0. I plan to pass the entire (immutable) job object to the target function.

Do you have Docker installed? The included docker-compose configuration provides an easy way to launch the various databases as containers simply by doing `docker-compose up`.

agronholm:
The problem with the tests does not seem to be related to the job stores though.

anon-john:
> I've hatched a more ambitious plan to implement this in APScheduler 4.0. I plan to pass the entire (immutable) job object to the target function.

do you plan to have another minor release before this (i.e. any chance to have the functionality of this PR before 4.0 is out)? 

One argument to put this in now is that it seems to me the logic you propose (passing of the immutable Job object) could be implemented based on this PR - rename the provide_scheduled_run_time add_job parameter, rename the scheduled_run_time func parameter and pass the Job object instead of the scheduled time to func (do I miss something here? apologies, if I'm wrong, this codebase is rather  new to me).

> Do you have Docker installed? The included docker-compose configuration provides an easy way to launch the various databases as containers simply by doing `docker-compose up`.

yes, tests I mentioned in the later comment ( https://github.com/agronholm/apscheduler/pull/380#issuecomment-507392611 ) were done with a complete setup (i.e. all dbs were set via docker-compose, no modifications were done to test_jobstores.py). 

> The problem with the tests does not seem to be related to the job stores though.

that was also my impression. I checked, that proposed PR is based on current master (it is). I've also tried merging it into the travistest branch, tests still run fine. Observed travis failures are even more surprising, if you note, that the problematic function is based on the test_add_live_job function, i.e. the call mechanics are identicall. The only non-trivial difference is in job function definition - lambda (test_add_live_job) versus local def (new test function). Should I make this function global and see if it helps (i.e. maybe this is some weird pickling related issue)? Anyway, please let me know if I could be of any more help here.

agronholm:
I ran the suite and the tests also fail for me locally, though not at the same point as on Travis. I can take a closer look later today.

I will definitely examine this PR when writing the job passing functionality, regardless of whether I choose to merge it.

richwifunds:
Getting the scheduled start time would be really helpful to a project I'm working on. We're using apscheduler to send jobs to a message broker rather than running the job directly. So we need to implement functionality like "misfire grace time" differently since a message can sit in the queue if there is no available consumer. We can do this by subtracting any time already passed since the scheduled run time from the misfire grace time and using that as the "time to expire" property of the message.

agronholm:
I plan to rearchitect APScheduler in a big way so that not only the start time but various other bits of information are (optinally) passed to the job.

anon-john:
Any news on this?

agronholm:
Yes, see the previous comment.

anon-john:
OK, so if I understand correctly - this won't be merged into 3.x series, this functionality should be available in 4.x, right? If so - is there any (rough) ETA on this?

agronholm:
I'm hoping to get this done before the end of August.

agronholm:
Closing since this will be implemented differently in 4.0.

PR #381, opened 2019-07-17
Crontab tweaks

kaos:
I ran into some issues where we need the crontab syntax to be more in line with the standard, so the output (actual run times for the job) aligns with what other tools say they would given a crontab expression.

This PR also addresses some cosmetics (adds support for `@macro` expressions as a short-hand for a full expression)

The two most notable fixes is support for day of week 0 being Sunday, and that when you specify both day of month _and_ day of week, you should run when either of them match, not both.

All changes should be transparent (fully backwards compatible). If you don't use "strict" mode, there is no changed behaviour.

### DIFF
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -91,7 +91,7 @@ class CronTrigger(BaseTrigger):
                 is_default = True
 
             field_class = self.FIELDS_MAP[field_name]
-            field = field_class(field_name, exprs, is_default)
+            field = field_class(field_name, exprs, is_default, standard=self.standard)
             self.fields.append(field)
 
     @classmethod
--- a/apscheduler/triggers/cron/__init__.py
+++ b/apscheduler/triggers/cron/__init__.py
@@ -43,6 +43,14 @@ class CronTrigger(BaseTrigger):
         'second': BaseField
     }
 
+    MACROS = {
+        '@yearly': '0 0 1 1 *',
+        '@monthly': '0 0 1 * *',
+        '@weekly': '0 0 * * 0',
+        '@daily': '0 0 * * *',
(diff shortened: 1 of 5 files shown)


agronholm:
I can implement these changes as the default in v4.0 if you give me a rundown of what exactly has changed in the logic. I'd rather not spend much time figuring that out from the changeset.

kaos:
Great. Most of it is actually already described in the relevant doc files. And examples are covered by the tests, the commit messages give a pretty good list of features, but here is a quick run down.

Everything relates to the arguments passed to the `from_crontab` function.

1. Macro syntaxes. such as `@daily`
2. Specifying a `?` for a field, has the same effect as leaving out the corresponding argument in the CronTrigger constructor.
3. When both day of month and day of week is specified, cron runs when either of the two matches, rather than when both matches as the current apscheduler implementation does.
4. Day 0 is Sunday.
5. Step interval support for day of week and months.
6. `L` for last day in month.
7. `Day#Occurence` in the day of week field. In apscheduler this is supported in the day of month field on the form `3rd wed`. I have a comment regarding this: https://github.com/agronholm/apscheduler/pull/381/files#diff-c05f03226892e46c7c82d5d48aef69edR136
8. Fixed a few regular expressions to support ranges also when using month/week names instead of numerals.

That should be most of it. As mentioned, all features introduced/changed are covered by the added tests. There aren't too many, so please have a quick look at them ;)

Also, I'd be happy to assist you with getting this into the 4.0 version, whenever you're ready for it, if you like.

candyapplecorn:
Thanks for writing this @kaos . It would fix the issues I've reported in https://github.com/agronholm/apscheduler/issues/495 .

andrewchen5678:
+1, just ran into the issue of day of the week 0 being Monday rather than Sunday and need this fix.

Issue #384, opened 2019-07-25
Standard tzinfo is not enough for apscheduler


b4stien:
## Expected Behavior

`apscheduler` should work with `next_run_time` specified with standard `tzinfo`.

## Current Behavior

`apscheduler` crashes.

## Steps to Reproduce

```
from datetime import datetime, timedelta
from dateutil.tz import tzutc

next_run_time = datetime.now(tzutc()) + timedelta(seconds=30)
scheduler.add_job(
    lambda: print("Hello"), trigger=IntervalTrigger(minutes=1), next_run_time=next_run_time
)

```

## Context (Environment)

`apscheduler` works only with `pytz` objects, the problematic call is there : https://github.com/agronholm/apscheduler/blob/18b50d9ee9ff14e816b557e34d7d3abc861d57e5/apscheduler/triggers/interval.py#L66

If `next_fire_time` is not stamped with a `pytz.tzinfo.BaseTzInfo`, it crashes.

`self.timezone` is a `pytz` object, `normalize()` is a `pytz` function which expects a `pytz.tzinfo.BaseTzInfo` as argument (see https://github.com/stub42/pytz/blob/master/src/pytz/tzinfo.py#L66) with an internal (and absent from the standard `tzinfo`) `_utcoffset` there : https://github.com/stub42/pytz/blob/master/src/pytz/tzinfo.py#L252.

agronholm:
Where did you get the idea that apscheduler should work with any tzinfo? The dependency on pytz timezones is mentioned in the documentation. This is specifically since only pytz timezones have the `localize()` and `normalize()` methods as you mentioned yourself.

b4stien:
From the signature there for instance : https://github.com/agronholm/apscheduler/blob/18b50d9ee9ff14e816b557e34d7d3abc861d57e5/apscheduler/triggers/interval.py#L22.

Also there : https://github.com/agronholm/apscheduler/blob/7e78cad90ce080825c46de8967d9573e9fc86353/apscheduler/job.py#L32

"It works with datetime, but if you specify a `tzinfo` it must comes from `pytz`" is not very explicit.

b4stien:
To precise my comment (and issue) the fact that when giving explicit timezone to apscheduler you have to use `pytz` stuff is not a problem (and I apologize for not reading the docs about it - tbh I don't use this feature).

My issue is that my standard datetimes with standard tzinfo does not work when given to `next_run_time`.

agronholm:
[...]

PR #391, opened 2019-09-17
add IntervalTrigger test with sting start_time

crhan:
Just to present the problem which I have met. But I cannot find a good way to fix it.


crhan:
I find out that this PR is relevant to #384

agronholm:
A pull request is hardly the right approach here when you don't have a solution. My solution in APScheduler 4.0 so far has been to remove the option of supplying dates as strings.

agronholm:
Your particular issue should be fixed in the 3.x branch now.

PR #393, opened 2019-09-23
Added seconds in cron schedular

aksandal:


agronholm:
This would break crontab for everyone. Why would the crontab compatibility mode even have to support seconds, when the cron standard does not, and the regular triggers constructor already supports seconds?

Issue #406, opened 2019-11-21
Refernce cycles in `executors.base_py3.run_coroutine_job`


chasezheng:
When a job raises an exception, thie method creates two reference cycles (See the graph below).

![refernce cycle](https://user-images.githubusercontent.com/16723977/69300276-8a01e980-0bc7-11ea-922e-48b6a042af93.png)

- The first reference cycle comes from the traceback in the local-frame. See the arrows, `f_locals`  and `tb`.
- The second reference cycle comes from the exception object in JobExecutionEvent

agronholm:
What tool did you use to generate this? It looks awesome!

agronholm:
That said, are you saying APScheduler causes a memory leak? If so, can you provide sample code that reproduces the problem?

chasezheng:
> What tool did you use to generate this? It looks awesome!

Haha I used [objgraph](https://mg.pov.lt/objgraph/) for this.



> That said, are you saying APScheduler causes a memory leak? If so, can you provide sample code that reproduces the problem?

Hopefully this will do. Note that at the end, after calling `run_coroutine_job`, the reference count will increment by 1.

```python
import sys
import asyncio
from datetime import datetime
from apscheduler.job import Job
from apscheduler.executors.base_py3 import run_coroutine_job

obj = object()

async def error_fn():
    # This function's local frame will likely not be garbage collected
    # Create a reference to obj.
    a = [obj]
    assert False
    
job = Job(scheduler=None)
job.func = error_fn
job.misfire_grace_time = 1000
job.name = 'mock'
job.args = ()
job.kwargs = {}
job.trigger = None

print(sys.getrefcount(obj))

asyncio.get_event_loop() \
    .run_until_complete(
         run_coroutine_job(
                job, 
                jobstore_alias=None, 
                run_times=[datetime.now().astimezone()],
                logger_name=__name__))

print(sys.getrefcount(obj))    # Count will increment by one
```

PR #415, opened 2020-01-11
Fixed reference cycle in run_coroutine_job when job fails

chasezheng:
For issue #406


agronholm:
I suggest that you make the PR against the 3.x branch. The master branch is still very much in flux due to the development of APScheduler 4.

PR #416, opened 2020-01-11
Ensured that jitter is always positive

### LINKED ISSUE #291: Cron jobs executes more times than desired when using jitter
When using the jitter option in a scheduled cron jobs, for example using the code from the documentation:
 
`# Run the `job_function` every sharp hour with an extra-delay picked randomly in a [-120,+120] seconds window.
sched.add_job(job_function, 'cron', hour='*', jitter=120)`

if the job is scheduled before the sharp hour then the job will be scheduled for the same hour again. For example:

- A Job is defined to be runned at sharp hours.

- Job is scheduled to run at 14:59:51

- Job runs at 14:59:51 but its scheduled to run at 15:00:02 again.
In this case we only want the job to run once at ~15:00:00.

Attached is a small piece of code to test this behaiviour: [test.zip](https://github.com/agronholm/apscheduler/files/1791019/test.zip)

agronholm:
Fixes #291.

### DIFF
--- a/apscheduler/triggers/base.py
+++ b/apscheduler/triggers/base.py
@@ -22,27 +22,16 @@ class BaseTrigger(six.with_metaclass(ABCMeta)):
 
     def _apply_jitter(self, next_fire_time, jitter, now):
         """
-        Randomize ``next_fire_time`` by adding or subtracting a random value (the jitter). If the
-        resulting datetime is in the past, returns the initial ``next_fire_time`` without jitter.
-
-        ``next_fire_time - jitter <= result <= next_fire_time + jitter``
+        Randomize ``next_fire_time`` by adding a random value (the jitter).
 
         :param datetime.datetime|None next_fire_time: next fire time without jitter applied. If
             ``None``, returns ``None``.
-        :param int|None jitter: maximum number of seconds to add or subtract to
-            ``next_fire_time``. If ``None`` or ``0``, returns ``next_fire_time``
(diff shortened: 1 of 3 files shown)


matslindh:
I'm testing the patch now - but from reading the code:

    seconds=random.uniform(-jitter, jitter)

is now:

    seconds=random.uniform(0, jitter)

Since the effective interval size from the jitter setting is now half what it used to be, at least a note should be added to the changelog (or possibly, to maintain compatibility with the old setting, `jitter*2`). Someone upgrading blindly could suddenly end up with their tasks concentrated in half the interval as before upgrading.

matslindh:
Patch works as expected outside of the narrowed jitter window. 

Since we now have a uniform jitter offset direction, we might want to add a check to see if `next run time + jitter` exceeds the originating next run time (i.e. if you have configured a cron scheduler to run every two minutes, but have a jitter larger than 120 seconds, you'll get skipped schedules).

It's an edge case and a configuration error, so I'm not sure how it'd be best to highlight it (if necessary at all).

Hendiadyoin1:
(I'm not the dev, just an idea)
Some people may want by directional jitter, so maybe add an option, where you can pass a slice, range, tuple(min,max) to offer more customisability

agronholm:
Sorry for taking so long with this.

matslindh:
No problem, thanks for the fix!

PR #417, opened 2020-01-12
Fixed reference cycle in run_coroutine_job

chasezheng:
For issue #406. This is using the code in apscheduler.executors.base.run_job.


chasezheng:
we don't need to check `if six.PY2` in a coroutine job. I will update the open another PR.

PR #418, opened 2020-01-12
Fixed reference cycle in run_coroutine_job

chasezheng:
For issue #406. This is using the code in apscheduler.executors.base.run_job.


agronholm:
Why do you keep opening new pull requests when you need to modify the code? Why not just push more commits?

agronholm:
I've pushed a solution in 1137ce1. Thank you for the PR anyway.

PR #424, opened 2020-03-22
Hours are 24 hour format

tc0nn:
Every monday at 2PM and 3PM would be 14 and 15 respectively.
I changed the example to include am and pm


agronholm:
Good catch, thanks!

PR #426, opened 2020-03-30
Jobstore with SQLite3

pascalziegler:
To avoid the dependency of sqlalchemy I wrote a jobstore with sqlite3. This could be used in https://github.com/borgbase/vorta/pull/377

### DIFF
--- a/apscheduler/jobstores/sqlite.py
+++ b/apscheduler/jobstores/sqlite.py
@@ -92,23 +92,19 @@ class SQLiteJobStore(BaseJobStore):
     def update_job(self, job):
         conn = sqlite3.connect(self.url)
         cursor = conn.cursor()
-        try:
-            with conn:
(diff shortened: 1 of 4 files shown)


agronholm:
I don't see why I should add a redundant job store just because someone doesn't want to install SQLAlchemy. Can you explain?

pascalziegler:
Well SQLite3 is contained in the standard python installation, so no additional packages are needed to use this jobstore. And I think there are more projects like Vorta which wants to keep the amount of dependencies small.

pascalziegler:
Do you know why the Error `fixture 'event_loop' not found` occurs? That Error was thrown even before I started to write that jobstore. I thought it was a problem with my local machine, but the last commit on master also failed on Travis.

agronholm:
> Well SQLite3 is contained in the standard python installation, so no additional packages are needed to use this jobstore.

APScheduler itself is not part of the standard library which kinda defeats that argument. Also, any new job stores I take on would increase the burden of maintenance and I don't see the benefit.

> And I think there are more projects like Vorta which wants to keep the amount of dependencies small.

I've never heard of this. Sorry to sound callous but why should I care?

PR #428, opened 2020-04-06
Add the ability to provide custom job class into the scheduler

ori-levi:


agronholm:
I'm sorry but I'm not accepting significant new features into the 3.x branch, and the master branch is currently undergoing major refactoring. When the first 4.0 pre-release version is out, check it out and tell me if you still need something like this.

PR #430, opened 2020-04-20
zookeeper plugin support windows

Laurel-rao:
function "os.path.join" return "\\" on windows, it can't work on zookeeper
replace os.path.join to "/", make apscheduler to support windows

### DIFF
--- a/apscheduler/jobstores/zookeeper.py
+++ b/apscheduler/jobstores/zookeeper.py
@@ -1,4 +1,3 @@
-import os
 import pickle
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


Laurel-rao:
- please check out the test/test_executor, it seems like event_loop can't be found.
```
file /home/travis/build/agronholm/apscheduler/tests/test_executors.py, line 150
  @pytest.fixture
  def asyncio_scheduler(event_loop):
E       fixture 'event_loop' not found
```
- checkout the Travis CI to see more information

agronholm:
I suggest you rebase this against the 3.x branch, as master is for 4.0 code and still in flux.

PR #431, opened 2020-04-21
Fix zookeeper plugin Bug on windows

Laurel-rao:
function "os.path.join" return "\" on windows, it can't work on zookeeper
replace os.path.join to "/", make apscheduler to support windows


Laurel-rao:
python 2.7 Can't be support, check travis CI information for more information

agronholm:
Looks good, thanks!

PR #436, opened 2020-06-03
add job_instance_id to scheduled job instance, for one job may has ma…

drunkpig:
this feature is useful when create project like django-apscheduler。
django-apscheduler has  serious problems when update job instance status, because django-apschedule don't konw which job instance status to update when there are more than one job instance running simultaneously.

### DIFF
--- a/apscheduler/events.py
+++ b/apscheduler/events.py
@@ -5,7 +5,6 @@ __all__ = ('EVENT_SCHEDULER_STARTED', 'EVENT_SCHEDULER_SHUTDOWN', 'EVENT_SCHEDUL
            'EVENT_JOB_ERROR', 'EVENT_JOB_MISSED', 'EVENT_JOB_SUBMITTED', 'EVENT_JOB_MAX_INSTANCES',
            'SchedulerEvent', 'JobEvent', 'JobExecutionEvent', 'JobSubmissionEvent')
 
-
 EVENT_SCHEDULER_STARTED = EVENT_SCHEDULER_START = 2 ** 0
 EVENT_SCHEDULER_SHUTDOWN = 2 ** 1
 EVENT_SCHEDULER_PAUSED = 2 ** 2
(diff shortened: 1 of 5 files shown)


agronholm:
Explain to me what a "job instance" is and how it differs from a "job".

drunkpig:
> Explain to me what a "job instance" is and how it differs from a "job".

```python

def class Job:
    def __init__(self, id,  name):
           pass


def class Jobinstance(Job):
    def __init__(self, id, name):
         super(id, name)
         self.instance_id = str(uuid.uuid4())

```

Job:  define the template of a task, But it not work until it is scheduled to run.

Jobinstance: A job can run in 3 times at the same time, we say it has 3 instance. Consider the parameter `max_instance`

One word: Job is a Class in OOP, Jobinstance is a class instance in OOP

agronholm:
This concept of a "job instance" seems alien to APScheduler. Why is there a secondary ID being created? For the record, I'm not going to accept this PR unless you can make a very strong case for it.

drunkpig:
> This concept of a "job instance" seems alien to APScheduler. Why is there a secondary ID being created? For the record, I'm not going to accept this PR unless you can make a very strong case for it.

let me show my code: https://github.com/drunkpig/django-apscheduler-ng

agronholm:
So am I correct in understanding that when a job gets queued on an executor, a new "job instance" is created?

drunkpig:
> So am I correct in understanding that when a job gets queued on an executor, a new "job instance" is created?

you're right!

drunkpig:
another strange problem is the event arriving to listener in a random order:  success event will arrive later than submission. this happen in the scenaro job has a very very shot exec time, success event will be send immediately, but other event a cached in a list.

so django-apscheduler update records in db in a mass. this is a very serious problem  cause that project in a failure.

I also suggest add timestamp in Event class. so the database record will have a precise tiem record about jobs.

agronholm:
> another strange problem is the event arriving to listener in a random order: success event will arrive later than submission. 

I don't see how this could work any other way. A job has to be submitted to an executor before its execution can succeed.

agronholm:
> I also suggest add timestamp in Event class. so the database record will have a precise tiem record about jobs.

The feature proposed in this PR is something that is already coming in APScheduler v4.0 (clean separation between task definitions, schedules and queued jobs), and I will not adding any new features to the 3.x branch. I will make sure the events are timestamped, though I have not yet decided if APScheduler should continue to provide its own event publishing system.

PR #442, opened 2020-06-30
Allow Job to be weak-referenced

bharel:
ATM, you can't weakref a job.
If a function in a job needs the job, it will cause circular reference.
Adding a `__weakref__` field will solve it.

### DIFF
--- a/apscheduler/job.py
+++ b/apscheduler/job.py
@@ -38,7 +38,7 @@ class Job:
 
     __slots__ = ('_scheduler', '_jobstore_alias', 'id', 'trigger', 'executor', 'func', 'func_ref',
                  'args', 'kwargs', 'name', 'misfire_grace_time', 'coalesce', 'max_instances',
-                 'next_run_time')
+                 'next_run_time', '__weakref__')
 
     def __init__(self, scheduler, id=None, **kwargs):
         super().__init__()
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -58,11 +58,13 @@ def test_remove(job):
     job.remove()
     job._scheduler.remove_job.assert_called_once_with(job.id, None)
 
+
 def test_weakref(create_job):
     job = create_job(func=dummyfunc)
     ref = weakref.ref(job)
     del job
     assert ref() is None
+
 
 def test_pending(job):
     """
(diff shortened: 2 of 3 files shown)


bharel:
Master is failing. None of the failures have anything to do with my PR.

agronholm:
What is the reason for this PR?

agronholm:
Master is a mess because I'm building APScheduler 4.0 in it and most of the architecture is not in place yet.

agronholm:
You're not supposed to store `Job` objects for long periods yourself. Store job ids instead and use them to remove the jobs from the scheduler when necessary.

agronholm:
Can't you just delete the `job` variable after you call `remove()`?

bharel:
> You're not supposed to store `Job` objects for long periods yourself. Store job ids instead and use them to remove the jobs from the scheduler when necessary.

The [https://apscheduler.readthedocs.io/en/stable/userguide.html#removing-jobs](user guide) suggests otherwise 🤷‍♂️ 

> Can't you just delete the `job` variable after you call `remove()`?

You're right, I can probably delete it from the closure using `nonlocal`, or by passing it as an argument but these are mostly workarounds for a field that is supposed to exist on almost every custom python object.

bharel:
Actually, if the argument is saved separately inside the job object, I won't be able to `del` it.
Also, if the job doesn't run, it won't be able to clear it's own variable from the closure, regardless of `nonlocal`, so the workaround isn't too promising 😢

agronholm:
The best I can do here is merge this into the v3.x branch. Master is still too much in the flux and the current `Job` class may be radically altered there or go away completely.

bharel:
Works for me 🤷 

Why will there be no job class? I think it's a fair design.

agronholm:
In APScheduler 4.0, jobs will be split into three parts:

1. Task definitions (named; contains the callable object along with several configuration parameters)
1. Schedules (always tied to task definitions)
1. Jobs (created by executors; always tied to task definitions but only optionally to schedules)

The name "job" on the last line is subject to change if deemed necessary.

agronholm:
I can make this change in 3.x but you need to rebase your code against the 3.x branch.

PR #449, opened 2020-07-26
Update faq.rst

pranavwagh:
This update in FAQ explains how can we implement parent child relationship between jobs without any code change.


pranavwagh:
It seems some checks failed, but It shouldn't matter for this pull request because this pull request isn't a code change

agronholm:
This is not a frequently asked question, and I am not inclined to add custom-made solutions to every imaginable use case, so I am rejecting this.

As a side note, the code style does not match the one in the scheduler or the rest of the documentation.

PR #451, opened 2020-08-04
remove a line

jiangzhuochi:
The line of "from apscheduler.scheduler import BlockingScheduler" is misleading. It doesn't seem to work correctly. 
Use "from apscheduler.schedulers.blocking import BlockingScheduler".


agronholm:
You're right, thanks!

Issue #466, opened 2020-09-30
Job with textual referenced function continually adds arg if using persistent job stores


spengjie:
## Sample Code to Reproduce
```python
from apscheduler.jobstores.mongodb import MongoDBJobStore
from apscheduler.schedulers.background import BackgroundScheduler

from myproject.config import mongodb_config


class DummyClass:

    def dummy_method(self, a, b):
        pass


instance = DummyClass()
jobstores = {
    'default': MongoDBJobStore(
        database='test',
        collection='scheduler_test',
        host=mongodb_config.connection_str,
    ),
}
scheduler = BackgroundScheduler(jobstores=jobstores)
job = scheduler.add_job('__main__:instance.dummy_method', args=(1, 2),
                        trigger='interval', seconds=1, id='1')

scheduler.start()
job = scheduler.reschedule_job('1', trigger='interval', seconds=2)
job = scheduler.reschedule_job('1', trigger='interval', seconds=2)
job = scheduler.reschedule_job('1', trigger='interval', seconds=2)
print(job.args)

while True:
    pass
```
## Expected Behavior
The value of `job.args` in the sample code should always be `()`.

## Current Behavior
The value of `job.args` in the sample code is `(<__main__.DummyClass object at 0x000001DDAEB5D828>, <__main__.DummyClass object at 0x000001DDAEB5D898>, <__main__.DummyClass object at 0x000001DDAEB5DBA8>)` when printing it. And its size is still being increasing if keep the script running.

## Environment
Python: 3.6.8
APScheduler: 3.6.3

## Context
The script will work fine if I change the add_job part to
```python
job = scheduler.add_job('__main__:DummyClass.dummy_method', args=(instance, 1, 2),
                        trigger='interval', seconds=1, id='1')
```
or
```python
job = scheduler.add_job(instance.dummy_method, args=(1, 2),
                        trigger='interval', seconds=1, id='1')
```
But I cannot use either of them. I'm using jsonrpc to add jobs remotely. So I have to use textual referenced function. And I'm using Celery to execute the job. Celery creates tasks from any callable by using a decorator. The decorator converts the callable to anther callable class instance. I cannot get the class instance.

## Detailed Description
I have located where the root cause is. Please see my analysis below.
[...]

PR #467, opened 2020-09-30
Fixed textual referenced bound methods issue on persistent job stores #466

spengjie:
For issue #466.

### DIFF
--- a/apscheduler/job.py
+++ b/apscheduler/job.py
@@ -242,8 +242,9 @@ class Job(object):
 
         # Instance methods cannot survive serialization as-is, so store the "self" argument
         # explicitly
-        if ismethod(self.func) and not isclass(self.func.__self__):
-            args = (self.func.__self__,) + tuple(self.args)
+        func = self.func
+        if ismethod(func) and not isclass(func.__self__) and not hasattr(func, '__func__'):
+            args = (func.__self__,) + tuple(self.args)
         else:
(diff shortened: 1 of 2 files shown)


agronholm:
Please also add a test that fails without this fix.

agronholm:
Did you notice that the test suite does not pass with this patch?

spengjie:
I'm on holiday, just got a chance to read the GitHub notifications. I‘ve tested locally without any issues. tox always stuck in PyQt5 installation in my local environment. So I didn't check the tox result. This is not right. I just fixed the tox problem, will check the test failures and add new test cases.

agronholm:
Looking good. The py3.5 test collection failure is probably just caused by test deps no longer supporting 3.5 because it has reached EOL.

spengjie:
In my local environment, tox always got InvocationError for all Python versions after I ran tox against py35 only. I deleted all the temp files and restored the code changes to restore the environment, but there's nothing changed. This is a very strange issue. I will continue the troubleshooting tomorrow.

agronholm:
Running pytest without tox works fine, but `tox -e py35` crashes with the same error as on Travis.

agronholm:
I've fixed the problem in the 3.x branch. Please rebase your code (or merge the latest changes)

agronholm:
Thanks!

PR #470, opened 2020-10-16
Allow Job to be weak-referenced

bharel:
Reupload of PR GH-442 after a rebase.

### DIFF
--- a/apscheduler/job.py
+++ b/apscheduler/job.py
@@ -40,7 +40,7 @@ class Job(object):
 
     __slots__ = ('_scheduler', '_jobstore_alias', 'id', 'trigger', 'executor', 'func', 'func_ref',
                  'args', 'kwargs', 'name', 'misfire_grace_time', 'coalesce', 'max_instances',
-                 'next_run_time')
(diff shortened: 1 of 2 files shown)


agronholm:
For future reference, you could've just rebased the code in the previous PR.

bharel:
Got stuck with lots of conflicts for some reason

agronholm:
The CI run is failing due to flake8 errors. If you want to test locally, you can start the necessary services with `docker-compose up -d` in the project directory and then run `tox`.

agronholm:
Your test relies on a reference counting garbage collector, and because pypy does not have one, the test fails. The suggested fix is to call `gc.collect()` at least once. I suggest you test this locally on pypy until it works.

bharel:
I thought of calling `gc.collect(1)`, just read about pypy's [garbage collector](https://doc.pypy.org/en/latest/cpython_differences.html#differences-related-to-garbage-collection-strategies). To be honest, I didn't think about that scenario. You learn a new thing every day :-)

I'm installing docker in the background for the tests.

agronholm:
Thanks!

PR #497, opened 2021-03-09
Fixed the incorrect doc on max_instances

Voyz:
max_instances is no longer set to 1, but to undefined - allowing an unlimited amount of concurrent job runs. Updated the documentation to reflect that default behaviour accurately.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -402,11 +402,12 @@ This is useful when you need to prune unwanted jobs before they have a chance to
 Limiting the number of concurrently executing instances of a job
 ----------------------------------------------------------------
 
-By default, only one instance of each job is allowed to be run at the same time.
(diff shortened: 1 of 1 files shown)


Voyz:
I don't understand what's going on but this PR is actually incorrect. 

* I started by having `max_instances` unset and with that setting an unlimited amount of job runs was allowed. 
* I then added `max_instances=1` and could accurately have the limiting to 1 concurrent job run
* I went ahead and submitted this PR thinking that indeed the docs are incorrect.
* Now I removed the `max_instances=1` and the jobs were still limited to only 1 concurrent run. 

I'm not using uWSGI, and I can't think what else could have been affecting that initial unlimited number of runs. Closing this PR in either way as this doesn't seem to be the case.

PR #503, opened 2021-04-01
Replace initialize() with start()

yoonghm:
`BlockingScheduler` has no attribute `initialize`.


agronholm:
There is no `BlockingScheduler` in master anymore. That particular change was a result of automated refactoring. I'm going to fix the examples before the first prerelease of v4.0.

PR #505, opened 2021-04-06
3.x

abood91:
changed the order of pyside2 imports


agronholm:
Why did the order of the imports have to change?

abood91:
the order should not matter, the reason behind my ordering is from newest framework to oldest

PR #518, opened 2021-06-19
Added the creation time of the job && MongoDBJobStore added create_time field

kwkwc:
Hi, first of all thanks to apscheduler.

Sometimes we want to know when a job was created, or we need to sort by creation time when getting all jobs directly from jobstores, so I added this property and the MongoDB field.

### DIFF
--- a/apscheduler/job.py
+++ b/apscheduler/job.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from inspect import ismethod, isclass
 from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


agronholm:
Thank you but I'm not currently taking PRs for new features, especially ones that cause disparity across the different job store implementations. You can of course keep these changes in your own customized job store class.

In v4.0 I will probably add such a field anyway, but that work will be done separately.

kwkwc:
Ok, thanks for your reply.

PR #533, opened 2021-07-28
Fixed / Removed Broken Example Links in Docs

williamhatcher:
I noticed the docs have links to example files which no longer exist. I removed those links and put a note in those files urging people to re-add or re-create example files.


agronholm:
What should be done here is link them to the `3.x` branch instead of master. The've only been deleted on master as part of work on v4.0.

williamhatcher:
@agronholm would you like me to do that and commit the changes?

agronholm:
I have fixed the links in the docs now.

PR #534, opened 2021-08-06
Outdated functions

vlad-outscraper:


agronholm:
The master branch is still work in progress for v4.0. I have not had the opportunity to update the examples yet. The RPC example in particular is no longer very relevant in v4.0 due to the addition of job store sharing.

PR #538, opened 2021-08-19
Updated URL

sasirajpuvvada:


agronholm:
There will be no "asyncio" scheduler in v4.0, but a generalized asynchronous scheduler and as such. The docs will be largely rewritten anyway, so I'm going to turn this down.

PR #546, opened 2021-09-09
Expose kwargs to underlying ThreadPool and ProcessPool executors

axuy:
Currently, there is no way to set some potentially important attributes in the underlying pool executors. This PR would allow a user to set these attributes.

Specifically, in my use case, I had an apscheduler app that used a lot of memory and delegated work to a ProcessPoolExecutor. Because the default mp_context for Unix is fork, this resulted in memory issues when the entire memory of my app was forked over to the new process.
https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

### DIFF
--- a/apscheduler/executors/pool.py
+++ b/apscheduler/executors/pool.py
@@ -44,12 +44,12 @@ class ThreadPoolExecutor(BasePoolExecutor):
     Plugin alias: ``threadpool``
 
     :param max_workers: the maximum number of spawned threads.
-    :param kwargs: dict of keyword arguments to pass to the underlying
+    :param pool_kwargs: dict of keyword arguments to pass to the underlying
         ThreadPoolExecutor constructor
     """
 
-    def __init__(self, max_workers=10, **kwargs):
-        pool = concurrent.futures.ThreadPoolExecutor(int(max_workers), **kwargs)
+    def __init__(self, max_workers=10, pool_kwargs={}):
+        pool = concurrent.futures.ThreadPoolExecutor(int(max_workers), **pool_kwargs)
         super(ThreadPoolExecutor, self).__init__(pool)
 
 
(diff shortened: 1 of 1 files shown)


apscheduler/executors/pool.py
```diff
@@ -44,10 +44,12 @@ class ThreadPoolExecutor(BasePoolExecutor):
     Plugin alias: ``threadpool``
 
     :param max_workers: the maximum number of spawned threads.
+    :param kwargs: dict of keyword arguments to pass to the underlying
+        ThreadPoolExecutor constructor
     """
 
-    def __init__(self, max_workers=10):
-        pool = concurrent.futures.ThreadPoolExecutor(int(max_workers))
+    def __init__(self, max_workers=10, **kwargs):
```
agronholm, line 51: I think I would prefer an actual dict parameter to be passed so that the job store can accept more parameters for itself too if deemed necessary.

agronholm (changes requested):
Just one change needed.

apscheduler/executors/pool.py
```diff
@@ -44,10 +44,12 @@ class ThreadPoolExecutor(BasePoolExecutor):
...
 
     :param max_workers: the maximum number of spawned threads.
+    :param pool_kwargs: dict of keyword arguments to pass to the underlying
+        ThreadPoolExecutor constructor
     """
 
-    def __init__(self, max_workers=10):
-        pool = concurrent.futures.ThreadPoolExecutor(int(max_workers))
+    def __init__(self, max_workers=10, pool_kwargs={}):
+        pool = concurrent.futures.ThreadPoolExecutor(int(max_workers), **pool_kwargs)
```
agronholm, line 52:
```suggestion
    def __init__(self, max_workers=10, pool_kwargs=None):
        pool_kwargs = pool_kwargs or {}
        pool = concurrent.futures.ThreadPoolExecutor(int(max_workers), **pool_kwargs)
```

apscheduler/executors/pool.py
```diff
@@ -58,8 +60,10 @@ class ProcessPoolExecutor(BasePoolExecutor):
...
 
     :param max_workers: the maximum number of spawned processes.
+    :param pool_kwargs: dict of keyword arguments to pass to the underlying
+        ProcessPoolExecutor constructor
     """
 
-    def __init__(self, max_workers=10):
-        pool = concurrent.futures.ProcessPoolExecutor(int(max_workers))
+    def __init__(self, max_workers=10, pool_kwargs={}):
+        pool = concurrent.futures.ProcessPoolExecutor(int(max_workers), **pool_kwargs)
```
agronholm, line 68:
```suggestion
    def __init__(self, max_workers=10, pool_kwargs=None):
        pool_kwargs = pool_kwargs or {}
        pool = concurrent.futures.ProcessPoolExecutor(int(max_workers), **pool_kwargs)
```

agronholm (changes requested):
While this code works, mutable default argument values are bad form and some code quality tools will complain about them. I've added a workaround as a suggestion.

agronholm (approved):

agronholm:
Thanks.

axuy:
> Thanks.

Thank you as well

PR #608, opened 2022-03-21
Fixed weekday order in CronTrigger

grillazz:
… the last value (7) is higher than the maximum value (6)"


agronholm:
You can't just make Sunday the first day of the week like that. Support of Sunday as 7 is not even universal.

agronholm:
That said, the documentation string should be updated since CronTrigger now follows the cron spec regarding which weekdays correspond to which numbers.

grillazz:
@agronholm i tested it carefully and 0 is mon and 6 is sun > this is how it is working now > so docs string should follow

agronholm:
Sure, but you didn't remove the last `sun` from the docstring. Do that and I will merge.

agronholm:
The title of the PR indicated that you wanted to make 7 = Sunday work with CronTrigger. But if you just want to make the docstring match the implementation, then that's all good.

grillazz:
PR updated

agronholm:
It seems that I was the one confused here – I didn't remember I had changed CronTrigger to work this way in master. Indeed 0 should mean Sunday and 6 should mean Saturday, to match with the crontab standard. If it doesn't work this way then it's a bug.

grillazz:
correct a here src/apscheduler/triggers/cron/expressions.py

```python
WEEKDAYS = ['mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun']

def get_weekday_index(weekday: str) -> int:
    try:
        return WEEKDAYS.index(weekday.lower())
    except ValueError:
        raise ValueError(f'Invalid weekday name {weekday!r}') from None
```

agronholm:
That is because Python considers 0 to be Monday: https://docs.python.org/3/library/datetime.html#datetime.date.weekday

agronholm:
It's not enough – `DayOfWeekField.get_value()` needs to be updated, along with the tests.

grillazz:
if you add this as bug I can add next PR

grillazz:
ok will do :)

grillazz:
@agronholm i think this PR is already doing fix https://github.com/agronholm/apscheduler/pull/381/files so maybe for now just update docs string to don't allow documentation confuse ppl

agronholm:
That PR adds unwanted things in the cron trigger. Simply fixing the weekday indexes would suffice.

grillazz:
ok thx will do this - just want to be sure that im not making duplicates

grillazz:
@agronholm pls review changes added as agreed

tests/triggers/test_cron.py
```diff
@@ -431,44 +431,44 @@ def test_year_list(timezone, serializer):
         (
             "* * * * *",
             "CronTrigger(year='*', month='*', day='*', week='*', day_of_week='*', "
-            "hour='*', minute='*', second='0', start_time='2020-05-19T19:53:22+02:00', "
+            "hour='*', minute='*', second='0', start_time='2020-05-19T19:53:22+02:00', ",
```
agronholm, line 434:
```suggestion
            "hour='*', minute='*', second='0', start_time='2020-05-19T19:53:22+02:00', "
```

PR #628, opened 2022-05-20
fix: Remove duplicate log

SeanGol:
When AsyncWorker is stopped normally, after the "Worker stopped" log, there is also a "Worker crashed" log, which should be removed it.


agronholm:
Good catch, thanks.

PR #641, opened 2022-07-06
Jitter can only be positive (per #416)

snqyz:


agronholm (changes requested):
Thanks for catching this, but I think the wording needs to be changed to be less confusing. Something about delaying the execution for a maximum of 120 seconds would be better.

agronholm:
Jitter has been moved out of trigger classes into schedules, so this PR is obsolete. Closing.

PR #669, opened 2022-10-14
quartz cron trigger attributes last, weekday, nth and single-word time shortcuts

christopherkoller:
Hi,

i tried to implement some features from the quartz cron trigger 

http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html

- L (“last”)  

> Has different meaning in each of the two fields in which it is allowed. 
> For example, the value “L” in the day-of-month field means “the last day of the month” - day 31 for January, day 28 for February on non-leap years. If used in the day-of-week field by itself, it simply means “7” or “SAT”.
> But if used in the day-of-week field after another value, it means “the last xxx day of the month” - for example “6L” means “the last friday of the month”. You can also specify an offset from the last day of the month, such as “L-3” which would mean the third-to-last day of the calendar month. When using the ‘L’ option, it is important not to specify lists, or ranges of values, as you’ll get confusing/unexpected results.

- W (“weekday”) 

> Used to specify the weekday (Monday-Friday) nearest the given day. As an example, if you were to specify “15W” as the value for the day-of-month field, the meaning is: “the nearest weekday to the 15th of the month”. 
> So if the 15th is a Saturday, the trigger will fire on Friday the 14th. 
> If the 15th is a Sunday, the trigger will fire on Monday the 16th. 
> If the 15th is a Tuesday, then it will fire on Tuesday the 15th. 
> However if you specify “1W” as the value for day-of-month, and the 1st is a Saturday, the trigger will fire on Monday the 3rd, 
> as it will not ‘jump’ over the boundary of a month’s days. The ‘W’ character can only be specified when the day-of-month is a single day, not a range or list of days.

- '# (nth)

> used to specify “the nth” XXX day of the month. 
> For example, the value of “6#3” in the day-of-week field means “the third Friday of the month” (day 6 = Friday and “#3” = the 3rd one in the month).
> Other examples: 
> “2#1” = the first Monday of the month and 
> “4#5” = the fifth Wednesday of the month. 
> Note that if you specify “#5” and there is not 5 of the given day-of-week in the month, then no firing will occur that month.

A good Cron expression generator - Quartz site to play around is

https://www.freeformatter.com/cron-expression-generator-quartz.html

I also added the single-word time shortcuts rules:

@daily: 0 0 0 * * * *
@midnight: 0 0 0 * * * *
@hourly: 0 0 * * * * *
@weekly: 0 0 0 * * 0 *
@monthly: 0 0 0 1 * * *
@annually: 0 0 0 1 1 * *
@yearly: 0 0 0 1 1 * *

I don't know if opening a pull request directly  is the best idea, so let me know what I should do to get the implementation accepted.

Best regards 

Chris


agronholm:
This looks interesting but it's a huge PR. Realistically I can't merge this until after the 4.0 final release.

agronholm:
In general, the less disruptive the changes are, the more likely I am to accept the PR. Completely refactoring the code for no good reason is harmful when trying to get your changes merged.

christopherkoller:
Thanks for the feedback!
My first idea was to not replace the original cron trigger, but adding a second one. But thats problably not a good solution.

I'm planning to write more tests to increase the code coverage and then start refactoring to fit your expectations.

I'm a bit surprised about all the tests failing, because on my local machine they're all working fine. ;-) 
Do you have any suggestions?

agronholm:
> I'm a bit surprised about all the tests failing, because on my local machine they're all working fine. ;-)

How did you run the tests? I get the exact same errors when running the test suite on your branch.

agronholm:
Doesn't look like this is going anywhere, so I'm closing.

PR #686, opened 2022-12-18
fix(docs): minor spelling edits in userguide

evanstjabadi:
I was reading the `userguide` docs and came across the `whixh`. Here is a quick fix along with 2 other minor edits


agronholm:
Thanks.

PR #701, opened 2023-01-16
add support for max_running_jobs in add_schedule method

ohayak:
max_running_jobs was missing from add_schedule methode


agronholm:
Thanks!

PR #704, opened 2023-01-27
SQLAlchemy 2.0 support

cfytrok:


agronholm:
The biggest problem is how to detect the presence of `RETURNING` support.

PR #726, opened 2023-04-02
[3.x] Fix pkg_resources deprecation - use importlib.metadata when available

bluetech:
Usage of `pkg_resources` now emits a deprecation warning (see [setuptools](https://github.com/pypa/setuptools/pull/3843)). It is also quite slow (just importing it).

Use `importlib.metadata` instead which is the recommended alternative.

### DIFF
--- a/apscheduler/schedulers/base.py
+++ b/apscheduler/schedulers/base.py
@@ -7,7 +7,6 @@ from logging import getLogger
 import warnings
 import sys
 
-from pkg_resources import iter_entry_points
 from tzlocal import get_localzone
 import six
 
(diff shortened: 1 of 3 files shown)


agronholm:
It would be better to use [importlib-metadata](https://pypi.org/project/importlib-metadata/) on Pythons earlier than 3.8.

apscheduler/schedulers/base.py
```diff
@@ -62,12 +66,18 @@ class BaseScheduler(six.with_metaclass(ABCMeta)):
...
     """
+    # The `group=...` API is only available in the backport, used in <=3.7, and in std>=3.10.
+    if (3, 8) <= sys.version_info <= (3, 9):
+        _trigger_plugins = {ep.name: ep for ep in entry_points()['apscheduler.triggers']}
+        _executor_plugins = {ep.name: ep for ep in entry_points()['apscheduler.executors']}
+        _jobstore_plugins = {ep.name: ep for ep in entry_points()['apscheduler.jobstores']}
+    else:
+        _trigger_plugins = {ep.name: ep for ep in entry_points(group='apscheduler.triggers')}
+        _executor_plugins = {ep.name: ep for ep in entry_points(group='apscheduler.executors')}
+        _jobstore_plugins = {ep.name: ep for ep in entry_points(group='apscheduler.jobstores')}
```
agronholm, line 77: On second thought, maybe the backport should be used for Pythons earlier than 3.10 since 3.10 is the first version where the API is stable (and `group` can be used uniformly).

bluetech: I still have a project on 3.9, and would prefer not to add an extra dependency for this somewhat minor detail. But let me know if you still want me to change it.

agronholm: Mhm...I don't have a strong opinion, given that this is a maintenance-only series. Unless a problem comes up, I won't insist.

PR #727, opened 2023-04-02
fix(sqlalchemy) - testing 'update returning' was raising error on SQLite

legobrick:
SET condition of UPDATE statement was not explicit. SQLite was not working.

Background:
```
2023-04-02T12:37:16+0200 PID:54053 {sqlalchemy.engine.Engine:2672} INFO - BEGIN (implicit)
2023-04-02T12:37:18+0200 PID:54053 {sqlalchemy.engine.Engine:1842} INFO - UPDATE metadata SET  RETURNING schema_version
2023-04-02T12:37:18+0200 PID:54053 {sqlalchemy.engine.Engine:1842} INFO - [generated in 0.00042s] ()
2023-04-02T12:37:20+0200 PID:54053 {sqlalchemy.engine.Engine:2675} INFO - ROLLBACK
2023-04-02T12:57:21+0200 PID:54053 {business.scheduling:61} ERROR - Unable to Create Schedule Object (sqlite3.OperationalError) near "RETURNING": syntax error [SQL: UPDATE metadata SET  RETURNING schema_version]
```

### DIFF
--- a/src/apscheduler/datastores/sqlalchemy.py
+++ b/src/apscheduler/datastores/sqlalchemy.py
@@ -321,8 +321,10 @@ class SQLAlchemyDataStore(BaseExternalDataStore):
         # Find out if the dialect supports UPDATE...RETURNING
         async for attempt in self._retry():
             with attempt:
-                update = self.t_metadata.update().values(schema_version=1).returning(
-                    self.t_metadata.c.schema_version
+                update = (
+                    self.t_metadata.update()
+                    .values(schema_version=1)
+                    .returning(self.t_metadata.c.schema_version)
                 )
                 async with self._begin_transaction() as conn:
                     try:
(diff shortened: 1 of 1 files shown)


agronholm:
So how is it that the tests pass with the sqlite store then? What am I missing?

legobrick:
> So how is it that the tests pass with the sqlite store then? What am I missing?

I'm using SQLAlchemy ^2.0.0 and aiosqlite. I don't know what's the cause between the 2 but they're my main suspects

agronholm:
Could you try with plain sqlite? For APScheduler, there is no advantage in using aiosqlite.

legobrick:
In the project I'm using APScheduler in, I also use SQLAlchemy + aiosqlite for other purposes. I can test that, but it wouldn't be much more than an experiment. Definitely not a realistic use case.

agronholm:
Yeah, I'm not asking you to make a permanent switch. I just want to narrow down the cause of the problem you're seeing.

agronholm:
I fixed this separately along with other SQLAlchemy related changes.

Issue #763, opened 2023-07-30
MongoDBDataStore causes future jobs to be run immediately


jiamingc:
### Things to check first

- [X] I have checked that my issue does not already have a solution in the [FAQ](https://apscheduler.readthedocs.io/en/master/faq.html)

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Version

master (9a93d1d9b3fac278dd558903359bc0b7f0c6ea14)

### What happened?

When I use `MongoDBDataStore` with `AsyncScheduler`, the scheduler doesn't wait until the correct time to run jobs. Here's a minimal reproduction of the issue:
```py
import asyncio
import sys
import logging
from datetime import datetime, timedelta, timezone
from pymongo import MongoClient
from apscheduler.schedulers.async_ import AsyncScheduler
from apscheduler.datastores.mongodb import MongoDBDataStore
from apscheduler.triggers.date import DateTrigger


async def job():
    logging.info(f"Running job")


async def main():
    async with AsyncScheduler(
        data_store=MongoDBDataStore(MongoClient("mongodb://root:password@localhost/"))
    ) as scheduler:
        await scheduler.add_schedule(
            job,
            DateTrigger(datetime.now(timezone.utc) + timedelta(seconds=30)),
        )
        await scheduler.run_until_stopped()


if __name__ == "__main__":
    logging.Formatter.converter = lambda *_: datetime.now(tz=timezone.utc).timetuple()
    logging.basicConfig(
        stream=sys.stdout,
        level=logging.INFO,
        format="%(asctime)s %(levelname)-4s %(name)s: %(message)s",
    )

    asyncio.run(main())
```
Resulting logs:
```
2023-07-30 20:11:02,216 INFO apscheduler.schedulers.async_: Added new schedule (task=Task(id='__main__:job', func=<function job at 0x7fd0a1e2ee80>, executor='async', max_running_jobs=None, misfire_grace_time=None, state=None), trigger=DateTrigger('2023-07-30 20:11:32.213015+00:00')); next run time at 2023-07-30 20:11:32.213015+00:00
2023-07-30 20:11:02,216 INFO apscheduler.schedulers.async_: Scheduler started
2023-07-30 20:11:02,227 INFO root: Running job
2023-07-30 20:11:02,228 INFO apscheduler.schedulers.async_: Job d6f39286-68c1-475c-8848-d230323051ee completed successfully
```

[...]

PR #768, opened 2023-08-05
Fix MongoDBDataStore bugs (#763)

jiamingc:
The reproduction from that issue works correctly after this patch.

### DIFF
--- a/src/apscheduler/datastores/mongodb.py
+++ b/src/apscheduler/datastores/mongodb.py
@@ -261,13 +261,14 @@ class MongoDBDataStore(BaseExternalDataStore):
         async for attempt in self._retry():
             with attempt, self.client.start_session() as session:
                 schedules: list[Schedule] = []
+                now = datetime.now(timezone.utc)
                 cursor = (
                     self._schedules.find(
                         {
-                            "next_fire_time": {"$ne": None},
+                            "next_fire_time": {"$lte": now},
                             "$or": [
                                 {"acquired_until": {"$exists": False}},
-                                {"acquired_until": {"$lt": datetime.now(timezone.utc)}},
+                                {"acquired_until": {"$lt": now}},
                             ],
                         },
                         session=session,
@@ -282,9 +283,7 @@ class MongoDBDataStore(BaseExternalDataStore):
 
                 if schedules:
                     now = datetime.now(timezone.utc)
-                    acquired_until = datetime.fromtimestamp(
-                        now.timestamp() + self.lock_expiration_delay, now.tzinfo
-                    )
+                    acquired_until = now + timedelta(seconds=self.lock_expiration_delay)
                     filters = {"_id": {"$in": [schedule.id for schedule in schedules]}}
                     update = {
                         "$set": {
@@ -354,13 +353,13 @@ class MongoDBDataStore(BaseExternalDataStore):
         async for attempt in self._retry():
             with attempt:
                 document = self._schedules.find_one(
-                    {"next_run_time": {"$ne": None}},
-                    projection=["next_run_time"],
-                    sort=[("next_run_time", ASCENDING)],
+                    {"next_fire_time": {"$ne": None}},
+                    projection=["next_fire_time"],
+                    sort=[("next_fire_time", ASCENDING)],
                 )
 
         if document:
-            return document["next_run_time"]
+            return document["next_fire_time"]
         else:
             return None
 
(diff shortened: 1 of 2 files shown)


agronholm:
There needs to be a test that fails without this patch to demonstrate that it fixes the issue (and prevents it from reoccurring later).

jiamingc:
I updated the datastore tests to test for the issues. I ran the tests on master to confirm that they catch the existing bugs:
```
================================================================== test session starts ==================================================================
platform linux -- Python 3.11.3, pytest-7.4.0, pluggy-1.2.0
rootdir: /home/fungus/workspace/apscheduler
configfile: pyproject.toml
plugins: cov-4.1.0, pytest_freezer-0.4.8, lazy-fixture-0.6.3, mock-3.11.1, anyio-4.0.0rc1
collected 95 items                                                                                                                                      

tests/test_datastores.py ........................F.....................................................................F                          [100%]

======================================================================= FAILURES ========================================================================
________________________________________________________ test_acquire_release_schedules[mongodb] ________________________________________________________
tests/test_datastores.py:198: in test_acquire_release_schedules
    assert not schedules3
E   AssertionError: assert not [Schedule(id='s3', task_id='task1', trigger=DateTrigger('2020-09-15 00:00:00+00:00'), args=(), kwargs={}, coalesce=<Co...i
nfo=<bson.tz_util.FixedOffset object at 0x7fa473396450>), last_fire_time=None, acquired_by=None, acquired_until=None)]
_________________________________________________________ test_next_schedule_run_time[mongodb] __________________________________________________________
tests/test_datastores.py:521: in test_next_schedule_run_time
    assert next_schedule_run_time == datetime(2020, 9, 13, tzinfo=timezone.utc)
E   assert None == datetime.datetime(2020, 9, 13, 0, 0, tzinfo=datetime.timezone.utc)
E    +  where datetime.datetime(2020, 9, 13, 0, 0, tzinfo=datetime.timezone.utc) = datetime(2020, 9, 13, tzinfo=datetime.timezone.utc)
E    +    where datetime.timezone.utc = timezone.utc
============================================================= 2 failed, 93 passed in 17.05s =============================================================
```

agronholm:
Thanks. I'll review this later today or tomorrow.

src/apscheduler/datastores/mongodb.py
```diff
@@ -354,15 +354,17 @@ async def get_next_schedule_run_time(self) -> datetime | None:
...
                 )
 
-        if document:
-            return document["next_run_time"]
-        else:
-            return None
+            if document:
+                return document["next_fire_time"]
+            else:
+                return None
```
agronholm, line 365: Why did you indent this code here?

src/apscheduler/datastores/mongodb.py
```diff
@@ -294,7 +292,9 @@ async def acquire_schedules(self, scheduler_id: str, limit: int) -> list[Schedul
                     }
                     self._schedules.update_many(filters, update, session=session)
 
-        return schedules
+            return schedules
```
agronholm, line 295: Why indent this code?

src/apscheduler/datastores/mongodb.py
```diff
@@ -281,10 +283,6 @@ async def acquire_schedules(self, scheduler_id: str, limit: int) -> list[Schedul
                     schedules.append(schedule)
 
                 if schedules:
-                    now = datetime.now(timezone.utc)
-                    acquired_until = datetime.fromtimestamp(
-                        now.timestamp() + self.lock_expiration_delay, now.tzinfo
-                    )
```
agronholm, line 287: Why remove this block? The query might take some time, and this code was there to take that into account.

agronholm (commented):
I agree with the fundamental two fixes ($ne -> $lte, next_run_time -> next_fire_time), but I have some questions.

src/apscheduler/datastores/mongodb.py
jiamingc: In the original, document is potentially undefined depending on how the replays play out. I'll revert it since this is probably better addressed separately.

src/apscheduler/datastores/mongodb.py
jiamingc: Reverting, but this was the same idea as below.

src/apscheduler/datastores/mongodb.py
jiamingc: Makes sense. Reverted.

src/apscheduler/datastores/mongodb.py
agronholm: If retries are exhausted, an exception should be raised, so the document would always be defined, yes?

agronholm:
Thanks!

src/apscheduler/datastores/mongodb.py
jiamingc: Ah, I see, makes sense.

PR #777, opened 2023-08-17
etcd jobstore support for apscheduler 3.x

Yan-Daojiang:
Recently, many people use etcd as their metadata hub, so I tried adding etcd persistent job storage for Apscheduler 3.x. I added the corresponding unit tests with 100% coverage, and they passed the tests locally.

### DIFF
--- a/apscheduler/jobstores/etcd.py
+++ b/apscheduler/jobstores/etcd.py
@@ -155,7 +155,7 @@ class EtcdJobStore(BaseJobStore):
                 failed_job_ids.append(failed_id)
                 self._logger.exception('Unable to restore job "%s" -- removing it', failed_id)
 
-        if failed_job_ids: 
+        if failed_job_ids:
             for failed_id in failed_job_ids:
                 self.remove_job(failed_id)
         paused_sort_key = datetime(9999, 12, 31, tzinfo=utc)
--- a/apscheduler/jobstores/etcd.py
+++ b/apscheduler/jobstores/etcd.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-
 from datetime import datetime
 
 from pytz import utc
--- a/apscheduler/jobstores/etcd.py
+++ b/apscheduler/jobstores/etcd.py
@@ -33,6 +33,7 @@ class EtcdJobStore(BaseJobStore):
     :param int pickle_protocol: pickle protocol level to use (for serialization), defaults to the
         highest available
     """
+
     def __init__(self, path='/apscheduler', client=None, close_connection_on_exit=False,
                  pickle_protocol=pickle.DEFAULT_PROTOCOL, **connect_args):
         super(EtcdJobStore, self).__init__()
--- a/apscheduler/jobstores/etcd.py
+++ b/apscheduler/jobstores/etcd.py
@@ -1,22 +1,18 @@
+import pickle
 from datetime import datetime
 
 from pytz import utc
-
-try:
-    import cPickle as pickle
-except ImportError:  # pragma: nocover
-    import pickle
-
-try:
-    from etcd3 import Etcd3Client
-except ImportError:  # pragma: nocover
-    raise ImportError('EtcdJobStore requires etcd3 be installed')
 
 from apscheduler.job import Job
 from apscheduler.jobstores.base import (BaseJobStore, ConflictingIdError,
                                         JobLookupError)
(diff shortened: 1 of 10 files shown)


apscheduler/jobstores/etcd.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Plugin alias: ``etcd``
+
+    :param str path: path to store jobs in
+    :param client: a :class:`~etcd3.client.etcd3` instance to use instead of
+        providing connection arguments
+    :param int pickle_protocol: pickle protocol level to use (for serialization), defaults to the
+        highest available
+    """
+    def __init__(self, path='/apscheduler', client=None, close_connection_on_exit=False,
```
agronholm, line 38:
```suggestion
    """

    def __init__(self, path='/apscheduler', client=None, close_connection_on_exit=False,
```

docs/contributing.rst
```diff
@@ -31,6 +31,7 @@ To fully run the test suite, you will need at least:
  * A MongoDB server
  * A Redis server
  * A Zookeeper server
+ * A Etcd server
```
agronholm, line 34:
```suggestion
 * An Etcd server
```

setup.py
```diff
@@ -53,6 +52,7 @@
         'tornado': ['tornado >= 4.3'],
         'twisted': ['twisted'],
         'zookeeper': ['kazoo'],
+        'etcd': ['etcd3', 'protobuf==3.20.0'],
```
agronholm, line 55: Is it really necessary to pin protobuf to a specific release? Why?

apscheduler/jobstores/etcd.py
```diff
@@ -0,0 +1,167 @@
...
+
+try:
+    import cPickle as pickle
+except ImportError:  # pragma: nocover
+    import pickle
+
+try:
+    from etcd3 import Etcd3Client
+except ImportError:  # pragma: nocover
+    raise ImportError('EtcdJobStore requires etcd3 be installed')
```
agronholm, line 15: According to PEP 8, conditional imports should be ordered last.

agronholm (changes requested):
I wouldn't have entertained the idea of adding a new back-end to the 3.x series but this PR was exceptionally well prepared, so I'm giving it a pass. I left some minor comments which should be addressed before merging though.

setup.py
Yan-Daojiang: This may be because the protobuf [package](https://pypi.org/project/protobuf/) might be installed without your involvement. Etcd3 has a dependency on protobuf. Using the latest protobuf requires setting additional environment variables，and according to the error description, this might impact your program performance.  Someone has the same [issue](https://github.com/kragniz/python-etcd3/issues/2242) in the official repository of python-etcd3.

Yan-Daojiang: This may be because the protobuf [package](https://pypi.org/project/protobuf/) might be installed without your involvement. Etcd3 has a dependency on protobuf. Using the latest protobuf requires setting additional environment variables，and according to the error description, this might impact your program performance.  Someone has the same [issue](https://github.com/kragniz/python-etcd3/issues/2242) in the official repository of python-etcd3.

Yan-Daojiang:
Hi agronholm! Thank you very much for your review of my code and your valuable suggestions, I made some changes to them. 
I noticed that your focus may be on 4.x at the moment, and some important changes have been made. I'd love to be able to make some contributions on 4.x in the future. Since pre-release 4.x is not currently in production, I only contributed to 3.x.

setup.py
agronholm: The problem seems to be that the `python-etcd3` project has not received any commits in a year, and fixing this would just require regenerating the stubs with a newer protobuf release.

Anyway, why not just pin it to `protobuf < 4`? Is there something wrong with the latest 3.x release (3.20.3)?

Yan-Daojiang: Yeah，you are right. Just 3.20.x or lower. I have changed it.

Yan-Daojiang:
There seems to be a problem in the unit test. This should not be caused by the new back-end. Could you help to take a look?

agronholm:
> There seems to be a problem in executor. This should not be caused by the new back-end. Could you help to take a look? When runing the test case `test_broken_pool` , the default memory storage should be used.

`test_broken_pool` is known to be flaky, and there's an open issue about it.

agronholm (approved):

Yan-Daojiang:
OK, thank you! I have no other questions.

PR #824, opened 2023-12-01
Fix `CronTrigger` deserialization for `JSONSerializer`

tdzienniak:
I noticed that the field `_last_fire_time` in `CronTrigger` is not properly deserialized when `JSONSerializer` is used, which causes the following error:
image

I added a converter to ensure this field is always a valid `datetime`.


agronholm:
Thanks, good catch! Now I have to fix the same issue in `IntervalTrigger` and `CalendarIntervalTrigger` too...

PR #847, opened 2024-01-13
Enhanced typing of `subscribe()`

### LINKED ISSUE #846: Typing of scheduler subscribe() method
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

Propose using a bound typevar for `AsyncScheduler.subscribe()`:

```diff
diff --git a/src/apscheduler/_schedulers/async_.py b/src/apscheduler/_schedulers/async_.py
index 5363dd3..963e09d 100644
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -11,7 +11,7 @@ from functools import partial
 from inspect import isbuiltin, isclass, ismethod, ismodule
 from logging import Logger, getLogger
 from types import TracebackType
-from typing import Any, Callable, Iterable, Mapping, cast
+from typing import Any, Callable, Iterable, Mapping, TypeVar, cast
 from uuid import UUID, uuid4
 
 import anyio
@@ -72,6 +72,7 @@ _microsecond_delta = timedelta(microseconds=1)
 _zero_timedelta = timedelta()
 
 TaskType: TypeAlias = "Task | str | Callable"
+EventType = TypeVar("EventType", bound="Event")
 
 
 @attrs.define(eq=False)
@@ -216,8 +217,8 @@ class AsyncScheduler:
 
     def subscribe(
         self,
-        callback: Callable[[Event], Any],
-        event_types: type[Event] | Iterable[type[Event]] | None = None,
+        callback: Callable[[EventType], Any],
+        event_types: type[EventType] | Iterable[type[EventType]] | None = None,
         *,
         one_shot: bool = False,
         is_async: bool = True,
```

### Use case

Curerently, a subscription like this:

```py
scheduler = AsyncScheduler()

def handler(event: JobAcquired) -> None:
    pass

scheduler.subscribe(handler, JobAcquired)
```
Raises a type error:

```
error: Argument 1 to "subscribe" of "AsyncScheduler" has incompatible type "Callable[[JobAcquired], None]"; expected "Callable[[Event], Any]"  [arg-type]
```

With proposed changeset, there is no error.

This may have additional benefit where a handler that cannot handle the subscribed types will be picked up statically, e.g:

```py
scheduler = AsyncScheduler()

def handler(event: JobReleased) -> None:
    pass

scheduler.subscribe(handler, JobAcquired)
```

With above patch applied, mypy raises:

```
error: Argument 1 to "subscribe" of "AsyncScheduler" has incompatible type "Callable[[JobReleased], None]"; expected "Callable[[JobAcquired], Any]"  [arg-type]
```

peterschutt:
This PR adds type var `T_Event` and uses it to type the `subscribe()` interfaces.

Fixes an issue where subscribing a handler would raise a type error if the handler is typed to receive a subclass of `Event`.

Allows type checkers to verify that the handler passed to `subscribe()` can support the event types it is assigned to handle.

Closes #846

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,12 +54,12 @@ test = [
     "anyio[trio]",
     "asyncmy >= 0.2.5; python_implementation == 'CPython'",
     "coverage >= 7",
-    "paho-mqtt >= 2.0",
+    "paho-mqtt >= 1.5",
     "psycopg",
     "pymongo >= 4",
     "pymysql[rsa]",
     "PySide6 >= 6.6; python_implementation == 'CPython'",
-    "pytest >= 8.0",
+    "pytest >= 7.4",
     "pytest-lazy-fixtures",
     "pytest-mock",
     "time-machine >= 2.13.0; python_implementation == 'CPython'",
@@ -132,4 +132,7 @@ commands = pyright --verifytypes apscheduler
 [testenv:docs]
 extras = doc
 commands = sphinx-build -n docs build/sphinx
+
+[testenv:paho-mqtt-v1]
+deps = paho-mqtt~=1.5
 """
--- a/src/apscheduler/_events.py
+++ b/src/apscheduler/_events.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from datetime import datetime, timezone
 from functools import partial
 from traceback import format_tb
-from typing import Any
+from typing import Any, TypeVar
 from uuid import UUID
 
 import attrs
@@ -13,6 +13,8 @@ from ._converters import as_aware_datetime, as_enum, as_uuid
 from ._enums import JobOutcome
 from ._structures import Job, JobResult
 from ._utils import qualified_name
+
+T_Event = TypeVar("T_Event", bound="Event")
 
 
 @attrs.define(kw_only=True, frozen=True)
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -38,6 +38,7 @@ from .._events import (
     SchedulerStarted,
     SchedulerStopped,
     ScheduleUpdated,
+    T_Event,
 )
 from .._exceptions import (
     CallableLookupError,
@@ -216,8 +217,8 @@ class AsyncScheduler:
 
     def subscribe(
         self,
-        callback: Callable[[Event], Any],
-        event_types: type[Event] | Iterable[type[Event]] | None = None,
+        callback: Callable[[T_Event], Any],
+        event_types: type[T_Event] | Iterable[type[T_Event]] | None = None,
         *,
         one_shot: bool = False,
         is_async: bool = True,
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -223,8 +223,7 @@ class AsyncScheduler:
         *,
         one_shot: bool = ...,
         is_async: bool = ...,
-    ) -> Subscription:
-        ...
+    ) -> Subscription: ...
 
     @overload
     def subscribe(
@@ -234,8 +233,7 @@ class AsyncScheduler:
         *,
         one_shot: bool = False,
         is_async: bool = True,
-    ) -> Subscription:
-        ...
+    ) -> Subscription: ...
 
     def subscribe(
         self,
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -11,7 +11,7 @@ from functools import partial
 from inspect import isbuiltin, isclass, ismethod, ismodule
 from logging import Logger, getLogger
 from types import TracebackType
-from typing import Any, Callable, Iterable, Mapping, cast
+from typing import Any, Callable, Iterable, Mapping, cast, overload
 from uuid import UUID, uuid4
 
 import anyio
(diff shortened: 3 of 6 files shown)


peterschutt:
TLDR;
---

+1:

- change improves simple case of single event type subscription (no type narrowing required in callback).

+0:

- multiple event type subscription is not any worse than status quo in behavior (still need to perform type narrowing in callback)

-1:

- inconsistent inference of event type unions
- how long it took to investigate and explain this

So I suppose that leaves me -1.

---

> I have some doubts whether this will work with multiple event types (ie. unions). Have you checked that?

I did, this example has no type error:

```py
from examples import scheduler

from apscheduler import JobAcquired, JobReleased


def handler(event: JobAcquired | JobReleased) -> None:
    ...


scheduler.subscribe(handler, {JobAcquired, JobReleased})
```

`JobAcquired` and `JobReleased` were what I was working with at the time, however this was unfortunate because:

```py
reveal_type({JobAcquired, JobReleased})  # Revealed type is "builtins.set[builtins.type]"
```
mypy narrows these down to `type`, not even `Event`. At the time I assumed there was no type error because it was type safe.

Instead, using `TaskAdded` and `TaskUpdated` b/c they are the first two in the `_events` module:

```py
reveal_type({TaskAdded, TaskUpdated})
# Revealed type is "builtins.set[def (*, timestamp: Any =, task_id: builtins.str) -> apscheduler._events.DataStoreEvent]
```

For these, mypy reduces them to a set of callables that return a `DataStoreEvent` instance, so that is what the handler needs to receive:

```py
from examples import scheduler

from apscheduler import DataStoreEvent, TaskAdded, TaskUpdated


def handler(event: DataStoreEvent) -> None:
    ...


scheduler.subscribe(handler, {TaskAdded, TaskUpdated})
```
B/c `DataStoreEvent` doesn't carry any more information than `Event` - this isn't any better than status quo, but its not worse either.

It would be less surprising if it would allow:

```py
def handler(event: TaskAdded | TaskUpdated) -> None:
    ....
```

However, in any case, type narrowing is going to be required in a multi-event type callback.

agronholm:
Thanks for taking the time to investigate. I'll mull it over once more.

peterschutt:
This is the nicest I've been able to get the behavior (diff is against this branch, and only adjusts the async scheduler interface, other interfaces would have to be updated also):

```diff
diff --git a/src/apscheduler/_schedulers/async_.py b/src/apscheduler/_schedulers/async_.py
index 6c5ae94..b975c90 100644
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -11,7 +11,7 @@ from functools import partial
 from inspect import isbuiltin, isclass, ismethod, ismodule
 from logging import Logger, getLogger
 from types import TracebackType
-from typing import Any, Callable, Iterable, Mapping, cast
+from typing import Any, Callable, Iterable, Mapping, cast, overload
 from uuid import UUID, uuid4
 
 import anyio
@@ -215,10 +215,30 @@ class AsyncScheduler:
         await self.data_store.cleanup()
         self.logger.info("Cleaned up expired job results and finished schedules")
 
+    @overload
     def subscribe(
         self,
         callback: Callable[[T_Event], Any],
-        event_types: type[T_Event] | Iterable[type[T_Event]] | None = None,
+        event_types: type[T_Event],
+        **kwargs: Any,
+    ) -> Subscription:
+        ...
+
+    @overload
+    def subscribe(
+        self,
+        callback: Callable[[Event], Any],
+        event_types: Iterable[type[Event]],
+        *,
+        one_shot: bool = ...,
+        is_async: bool = ...,
+    ) -> Subscription:
+        ...
+
+    def subscribe(
+        self,
+        callback: Callable[[T_Event], Any] | Callable[[Event], Any],
+        event_types: type[T_Event] | Iterable[type[Event]] | None = None,
         *,
         one_shot: bool = False,
         is_async: bool = True,
```

With this pattern it allows the callback to share the type of a single event type subscription, for example, this type checks fine:

```py
def handler(event: JobAcquired) -> None:
    ...

scheduler.subscribe(handler, JobAcquired)
```
And for the iterable of types it behaves the same as it does now, i.e., this is an error:

```py
def handler(event: TaskAdded | TaskUpdated) -> None:
    ...

scheduler.subscribe(handler, {TaskAdded, TaskUpdated})
# error: Argument 1 to "subscribe" of "AsyncScheduler" has incompatible type "Callable[[TaskAdded | TaskUpdated], None]"; expected "Callable[[Event], Any]"  [arg-type]
```

This is no error:

```py

def handler(event: Event) -> None:
    ...

scheduler.subscribe(handler, {TaskAdded, TaskUpdated})
```

Its not a bad middle ground as it removes the need to have a redundant type narrowing or type ignore where the handler should only receive a single type, and leaves the rest of the behavior as it is.

We could go further and define overloads for `DataStoreEvent` and `SchedulerEvent` so that this would type check OK:

```py
def handler(event: DataStoreEvent) -> None:
    ...

scheduler.subscribe(handler, {TaskAdded, TaskUpdated})
```
But given that neither of those types define any attributes that are common across all of their sub types, it  wouldn't be of any benefit.

agronholm:
It sounds like it's at least an improvement over the existing typing. Any idea if the test failures are caused by these changes?

agronholm:
Hm, looks more like an incompatibility between pytest 8 and pytest-lazyfixture.

agronholm:
Yup, looks like: https://github.com/TvoroG/pytest-lazy-fixture/issues/65

agronholm:
Alright, finally got around to merging this. Thanks!

PR #865, opened 2024-02-16
tests: move to pytest-lazy-fixtures dep & mqtt update

peterschutt:
This PR addresses failing tests due to pytest-lazy-fixture incompatibility with pytest v8, and breaking changes in paho.mqtt v2.

Replaces pytest-lazy-fixture with pytest-lazy-fixtures, and adjusts use of mqtt client to conform to new api.

@agronholm now that https://github.com/dev-petrov/pytest-lazy-fixtures/issues/5 is closed and released I didn't know if you may want to jump over to that lib now to get things moving again. Either way, I can back that out and re-pin to pytest 7 and just address the mqtt stuff.


agronholm:
Not that it matters a lot, but why upgrade the pytest dependency?

tests/conftest.py
```diff
@@ -69,10 +69,11 @@ async def redis_broker(serializer: Serializer) -> EventBroker:
 @pytest.fixture
 def mqtt_broker(serializer: Serializer) -> EventBroker:
     from paho.mqtt.client import Client
+    from paho.mqtt.enums import CallbackAPIVersion
 
     from apscheduler.eventbrokers.mqtt import MQTTEventBroker
 
-    return MQTTEventBroker(Client(), serializer=serializer)
+    return MQTTEventBroker(Client(CallbackAPIVersion.VERSION1), serializer=serializer)
```
agronholm, line 76: If we want this to be forward compatible, we should use the v2 callback API. Agreed?

peterschutt: Have to admit I was just trying to get the tests to run. For proper back/forward compat I had to relax the signatures on the `MQTTEventBroker` callback methods so they can handle either v1 or v2 style clients, adjusted the `mqtt_client` pytest fixture to work with either 1.x and 2.x, and added a tox env to explicitly test against v1 (`$ tox -e . paho-mqtt-v1`). LMK thoughts on all that.

peterschutt:
> Not that it matters a lot, but why upgrade the pytest dependency?



> Not that it matters a lot, but why upgrade the pytest dependency?

Changed to `>=7.4`.

tests/conftest.py
agronholm: Do we really need to support paho-mqtt 1.x when there was no support for it in any previous stable release? I think not. Besides, that version check you're doing doesn't play nice with application packagers like PyInstaller which omit packaging metadata by default.

peterschutt: Thanks that simplifies it a lot. Latest change reverts my last commit and pins the mqtt dep to 2+. Changed the callback signatures per types [here](https://github.com/eclipse/paho.mqtt.python/blob/ebffdc2ca0a43fd87dfd270bbb16007591139626/src/paho/mqtt/client.py#L250-L277), but we could still relax them and be able to support `VERSION1` and `VERSION2` callbacks??

agronholm:
Thanks!

PR #889, opened 2024-04-08
SQLAlchemy store should remove finished schedules on cleanup

Reskov:
It looks like a typo error because a similar code in the mongodb/memory module deletes finished schedules. 

I have modified a test case for two reasons: 
1) it was not executed on all data stores 
2) the `anyio.Event()` is not serializable across different datastores, resulting in an error before execution.


agronholm:
Looks good at a glance. I'll give it a proper review within a couple days.

agronholm:
Thank you!

PR #891, opened 2024-04-11
Fix issue with sqlalchemy cleanup deleting all unfinished schedules instead of finished schedules

BraedonLeonard:


BraedonLeonard:
Closing, since it looks like someone else beat me to the punch a few days ago

PR #898, opened 2024-04-23
Avoid infinite loop in CalendarIntervalTrigger with UTC timezone

### LINKED ISSUE #897: Infinite loop in CalendarIntervalTrigger with UTC timezone
### Things to check first

### Version

4.0.0a4

### What happened?

Time with UTC timezone will be always not equal to other time.

```python
from datetime import time
from zoneinfo import ZoneInfo

utctime = time(0, 0, tzinfo=ZoneInfo('UTC'))

a = time(0, 0)
b = time(0, 0, tzinfo=ZoneInfo("Asia/Shanghai"))
c = time(0, 0, tzinfo=ZoneInfo("Asia/Tokyo"))

print(a == utctime)  # False
print(a == b == c)   # True
```

So the condition here will always be `False`

https://github.com/agronholm/apscheduler/blob/3e42a1bd73da947634ea94724e5221a43ad7ce0c/src/apscheduler/triggers/calendarinterval.py#L121

It should be call `.timetz()` to get the time part.

```python
from apscheduler.triggers.calendarinterval import CalendarIntervalTrigger

trigger = CalendarIntervalTrigger(days=1, timezone="UTC")

print(trigger.next())

  File "<stdin>", line 1, in <module>
  File "D:\.venv\Lib\site-packages\apscheduler\triggers\calendarinterval.py", line 109, in next
    next_time = datetime.fromtimestamp(timestamp, self.timezone)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 22] Invalid argument
```

unights:
Fixes #897

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -20,6 +20,7 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed JSON serialization of triggers that had been used at least once
 - Fixed dialect name checks in the SQLAlchemy job store
 - Fixed JSON and CBOR serializers unable to serialize enums
+- Fixed infinite loop in CalendarIntervalTrigger with UTC timezone
 
 **4.0.0a4**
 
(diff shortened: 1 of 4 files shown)


agronholm:
Ok, looks good. If you could also add a changelog note (`docs/versionhistory.rst`), then I'll merge.

agronholm:
Thank you!

PR #900, opened 2024-04-24
Add missing test dependency on `greenlet`

WillDaSilva:
Without this dependency installed, hundreds of tests fail or error when running `pytest` after following the instructions in `docs/contributing.rst`. I encountered this problem while testing the library under macOS - maybe it isn't a problem on other platforms.


agronholm:
The proper way to ensure that greenlet is installed is to install SQLAlchemy with the `asyncio` extra:
```toml
[project.optional-dependencies]
sqlalchemy = ["sqlalchemy[asyncio] >= 2.0.19"]
```

agronholm:
Ty!

Issue #899, opened 2024-04-23
Support pausing and unpausing schedules


WillDaSilva:
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

This feature would involve adding a `paused` boolean read-only property to the schedules. When `True`, the schedule runs as normal. When `False`, no jobs are created for the schedule, and the `next_fire_time` does not get updated`.

The queries made by APScheduler to the data stores to get the schedules that need to fire should be updated so that they do not return any schedules with `paused` set to `False`.

The following two methods could be added to the `Schedule` class:

```python
def pause(self) -> None:
    """Pause the schedule."""
    ...

def unpause(self, *, next_fire_time: datetime | None = None) -> None:
    """Unpause the schedule.

    Parameters:
        next_fire_time: The next time the schedule should fire after being unpaused, or
            `None` to compute the `next_fire_time` from the schedule's trigger.
    """
    ...
```

### Use case

This would be useful for any case where a user may want a schedule to not run for some indeterminate period of time. In my case APScheduler is used to manage/run schedules that are configured by the users of our platform. These users have expressed a desire to temporarily disable their schedules without fully deleting them, since they don't want to lose the information they configured in the trigger, or args/kwargs.

agronholm:
Schedule pausing functions need to be part of the scheduler, not the schedule object, as the schedule object does not hold a reference to the scheduler. Further, I don't think it's a good idea to force a specific next fire time. A better idea would be a flag that allowed the schedule to be resumed either from where it left, or from the current time.

WillDaSilva:
Thanks for the feedback @agronholm. I've updated the description to take it into account.

WillDaSilva:
I'm working on a PR for this, and I noticed that the majority of the work that is required is to support the configurable unpause behaviour. I think that as a first-pass it would be worthwhile to always resume the schedules from where they left off, and then if there is a desire for the ability to unpause the schedule such that its next fire time is as it would be had the schedule been created now, that feature can be added later on.

@agronholm When a schedule is unpaused, if it would have run had it not been paused, do you think we ought to trigger its normal misfire behaviour? That would be the easiest way to implement this feature, but I could imagine users wanting to be able to pause a schedule to skip some triggers, and then unpause without the misfire behaviour triggering.

Maybe as a first-pass we can make it trigger the misfire behaviour if any triggers were missed, and then if users want to be able to ignore misfires for paused schedules, that can be added in a future PR.

[...]

PR #902, opened 2024-05-02
feat: Support pausing and unpausing schedules

WillDaSilva:
Closes https://github.com/agronholm/apscheduler/issues/899

### DIFF
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -532,17 +532,17 @@ class AsyncScheduler:
         self._check_initialized()
         await self.data_store.remove_schedules({id})
 
-    async def pause_schedule(self, schedule_id: str) -> None:
+    async def pause_schedule(self, id: str) -> None:
         """Pause the specified schedule."""
         self._check_initialized()
         await self.data_store.add_schedule(
-            schedule=attrs.evolve(await self.get_schedule(schedule_id), paused=True),
+            schedule=attrs.evolve(await self.get_schedule(id), paused=True),
             conflict_policy=ConflictPolicy.replace,
         )
 
     async def unpause_schedule(
         self,
-        schedule_id: str,
+        id: str,
         *,
         resume_from: datetime | Literal["now"] | None = None,
     ) -> None:
@@ -556,7 +556,7 @@ class AsyncScheduler:
 
         """
         self._check_initialized()
-        schedule = await self.get_schedule(schedule_id)
+        schedule = await self.get_schedule(id)
 
         if resume_from is None:
             next_fire_time = schedule.next_fire_time
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -532,28 +532,63 @@ class AsyncScheduler:
         self._check_initialized()
         await self.data_store.remove_schedules({id})
 
-    async def pause_schedule(self, id: str) -> None:
+    async def pause_schedule(self, schedule_id: str) -> None:
         """Pause the specified schedule."""
         self._check_initialized()
-        await self.data_store.pause_schedules({id})
+        await self.data_store.add_schedule(
+            schedule=attrs.evolve(await self.get_schedule(schedule_id), paused=True),
+            conflict_policy=ConflictPolicy.replace,
+        )
+
+    def _get_unpaused_next_fire_time(
+        self,
+        schedule: Schedule,
+        resume_from: datetime | Literal["now"] | None,
+    ) -> datetime | None:
+        if resume_from is None:
+            return schedule.next_fire_time
+        if resume_from == "now":
+            resume_from = datetime.now(tz=timezone.utc)
+        if (
+            schedule.next_fire_time is not None
+            and schedule.next_fire_time >= resume_from
+        ):
+            return schedule.next_fire_time
+        try:
+            while (next_fire_time := schedule.trigger.next()) < resume_from:
+                pass  # Advance `next_fire_time` until its at or past `resume_from`
+        except TypeError:  # The trigger is exhausted
+            return None
+        return next_fire_time
 
     async def unpause_schedule(
         self,
-        id: str,
+        schedule_id: str,
         *,
         resume_from: datetime | Literal["now"] | None = None,
     ) -> None:
-        """Unpause the specified schedule.
+        """
+        Unpause the specified schedule.
 
-        By default, the schedule will be resumed as if it had never been paused, and all
-        missed runs will be considered misfires. The ``resume_from`` parameter can be
-        used to specify a different time from which to resume the schedule. The string
-        ``'now'`` can be used as shorthand for ``datetime.now(tz=UTC)``. If
-        ``resume_from`` is not ``None``, then the trigger will be repeatedly advanced
-        until the next fire time is at or after the specified time.
+
+        :param resume_from: the time to resume the schedules from, or ``'now'`` as a
+            shorthand for ``datetime.now(tz=UTC)`` or ``None`` to resume from where the
+            schedule left off which may cause it to misfire
+
         """
         self._check_initialized()
-        await self.data_store.unpause_schedules({id}, resume_from=resume_from)
+        schedule = await self.get_schedule(schedule_id)
+        await self.data_store.add_schedule(
+            schedule=attrs.evolve(
+                schedule,
+                paused=False,
+                next_fire_time=self._get_unpaused_next_fire_time(
+                    schedule,
+                    resume_from,
+                ),
+            ),
+            conflict_policy=ConflictPolicy.replace,
+        )
 
     async def add_job(
         self,
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -11,7 +11,7 @@ from functools import partial
 from inspect import isbuiltin, isclass, ismethod, ismodule
 from logging import Logger, getLogger
 from types import TracebackType
-from typing import Any, Callable, Iterable, Mapping, cast
+from typing import Any, Callable, Iterable, Mapping, cast, overload
 from uuid import UUID, uuid4
 
 import anyio
@@ -44,6 +44,7 @@ from .._events import (
     SchedulerStarted,
     SchedulerStopped,
     ScheduleUpdated,
+    T_Event,
 )
 from .._exceptions import (
     CallableLookupError,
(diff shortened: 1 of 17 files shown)


src/apscheduler/_schedulers/async_.py
```diff
@@ -30,7 +30,13 @@
...
 from .._context import current_async_scheduler, current_job
 from .._converters import as_enum, as_timedelta
-from .._enums import CoalescePolicy, ConflictPolicy, JobOutcome, RunState, SchedulerRole
+from .._enums import (
+    CoalescePolicy,
+    ConflictPolicy,
+    JobOutcome,
+    RunState,
+    SchedulerRole,
+)
```
peterschutt, line 39: I notice your PR changes the formatting of the imports in a couple of places, seemingly without a need. Just FYI.

WillDaSilva, line 39:
```suggestion
from .._enums import CoalescePolicy, ConflictPolicy, JobOutcome, RunState, SchedulerRole
```

src/apscheduler/_schedulers/sync.py
```diff
@@ -6,17 +6,22 @@
...
 from anyio.from_thread import BlockingPortal, start_blocking_portal
 
 from .. import current_scheduler
-from .._enums import CoalescePolicy, ConflictPolicy, RunState, SchedulerRole
+from .._enums import (
+    CoalescePolicy,
+    ConflictPolicy,
+    RunState,
+    SchedulerRole,
+)
```
WillDaSilva, line 24:
```suggestion
from .._enums import CoalescePolicy, ConflictPolicy, RunState, SchedulerRole
```

src/apscheduler/_schedulers/async_.py
WillDaSilva: Addressed in 695efe6

src/apscheduler/_schedulers/async_.py
```diff
@@ -529,6 +532,29 @@ async def remove_schedule(self, id: str) -> None:
         self._check_initialized()
         await self.data_store.remove_schedules({id})
 
+    async def pause_schedule(self, id: str) -> None:
```
agronholm, line 535: The `id` parameter shadows the built-in `id()` function. You should rename it to `schedule_id`.

src/apscheduler/_schedulers/async_.py
```diff
@@ -529,6 +532,29 @@ async def remove_schedule(self, id: str) -> None:
...
+        """Pause the specified schedule."""
+        self._check_initialized()
+        await self.data_store.pause_schedules({id})
```
agronholm, line 538: This code should be in the scheduler (i.e. this method) if it doesn't require any additional functionality from the data store.

src/apscheduler/_schedulers/async_.py
```diff
@@ -529,6 +532,29 @@ async def remove_schedule(self, id: str) -> None:
...
+
+        By default, the schedule will be resumed as if it had never been paused, and all
+        missed runs will be considered misfires. The ``resume_from`` parameter can be
+        used to specify a different time from which to resume the schedule. The string
+        ``'now'`` can be used as shorthand for ``datetime.now(tz=UTC)``. If
+        ``resume_from`` is not ``None``, then the trigger will be repeatedly advanced
+        until the next fire time is at or after the specified time.
+        """
+        self._check_initialized()
+        await self.data_store.unpause_schedules({id}, resume_from=resume_from)
```
agronholm, line 556: As above, this code should be in the scheduler, not the data store.

src/apscheduler/_schedulers/async_.py
```diff
@@ -529,6 +532,29 @@ async def remove_schedule(self, id: str) -> None:
...
+        resume_from: datetime | Literal["now"] | None = None,
+    ) -> None:
+        """Unpause the specified schedule.
+
+        By default, the schedule will be resumed as if it had never been paused, and all
+        missed runs will be considered misfires. The ``resume_from`` parameter can be
+        used to specify a different time from which to resume the schedule. The string
+        ``'now'`` can be used as shorthand for ``datetime.now(tz=UTC)``. If
+        ``resume_from`` is not ``None``, then the trigger will be repeatedly advanced
+        until the next fire time is at or after the specified time.
```
agronholm, line 553: This is missing Sphinx-style parameter documentations (see other methods for examples). Also, the convention, at least in my projects, is to start the docstring text on a new line after the triple quotes, and add an empty line at the end if the docstring is immediately followed by code.

src/apscheduler/datastores/mongodb.py
```diff
@@ -366,9 +366,19 @@ async def acquire_schedules(self, scheduler_id: str, limit: int) -> list[Schedul
...
+                                        {"paused": False},
+                                    ]
+                                },
+                                {
+                                    "$or": [
+                                        {"acquired_until": {"$exists": False}},
+                                        {"acquired_until": {"$lt": now}},
+                                    ]
+                                },
                             ],
```
agronholm, line 382: Do we know if mongo can figure out that it should use the index for `acquired_until`, or do we need to give it some kind of a hint?

agronholm (changes requested):
Nice job. I'll need some changes though.

src/apscheduler/datastores/mongodb.py
agronholm: Oh, never mind, we don't have any index on `acquired_until` either.

src/apscheduler/_schedulers/async_.py
WillDaSilva: @agronholm For this I was just following what was used by other functions, since consistency is valuable for parts of the external API like this. Would you still rather I use `schedule_id` even though the other similar methods use `id`?

src/apscheduler/_schedulers/async_.py
WillDaSilva: Addressed in 61f7051

src/apscheduler/_schedulers/async_.py
WillDaSilva: Addressed in 61f7051

src/apscheduler/_schedulers/async_.py
WillDaSilva: Good call. Addressed in 61f7051. One consequence of this is that it no longer makes sense to have datastore tests for pausing and unpausing, so those tests have been removed.

src/apscheduler/_schedulers/async_.py
WillDaSilva: In 61f7051 I've updated `pause_schedule` and `unpause_schedule` to use `schedule_id` instead of `id`. If you'd rather `id` be used consistently, I can revert that change.

src/apscheduler/datastores/sqlalchemy.py
```diff
@@ -752,7 +755,10 @@ async def get_next_schedule_run_time(self) -> datetime | None:
 
         statenent = (
             select(*columns)
-            .where(self._t_schedules.c.next_fire_time.isnot(None))
+            .where(
+                self._t_schedules.c.next_fire_time.isnot(None),
+                self._t_schedules.c.paused.is_(False),
```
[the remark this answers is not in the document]
WillDaSilva: Addressed in 61f7051

src/apscheduler/datastores/sqlalchemy.py
```diff
@@ -600,6 +602,7 @@ async def acquire_schedules(self, scheduler_id: str, limit: int) -> list[Schedul
                             and_(
                                 self._t_schedules.c.next_fire_time.isnot(None),
                                 comparison,
+                                self._t_schedules.c.paused.is_(False),
```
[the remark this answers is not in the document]
WillDaSilva: Addressed in 61f7051

src/apscheduler/datastores/memory.py
```diff
@@ -207,8 +207,11 @@ async def acquire_schedules(self, scheduler_id: str, limit: int) -> list[Schedul
         schedules: list[Schedule] = []
         for state in self._schedules:
             if state.next_fire_time is None or state.next_fire_time > now:
-                # The schedule is either paused or not yet due
-                break
+                # The schedule is either exhausted or not yet due
+                continue
```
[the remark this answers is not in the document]
WillDaSilva: Ah right, I was mistakenly under the impression that other schedules past this point could be due. I've reverted this to use `break` instead of `continue`, and left a clarifying comment to hopefully help other contributors in the future not make my mistake.

src/apscheduler/_schedulers/async_.py
```diff
@@ -529,6 +532,64 @@ async def remove_schedule(self, id: str) -> None:
...
 
+    async def pause_schedule(self, schedule_id: str) -> None:
+        """Pause the specified schedule."""
+        self._check_initialized()
+        await self.data_store.add_schedule(
+            schedule=attrs.evolve(await self.get_schedule(schedule_id), paused=True),
+            conflict_policy=ConflictPolicy.replace,
+        )
+
+    def _get_unpaused_next_fire_time(
```
agronholm, line 543: This method doesn't seem to be used anywhere but `unpause_schedule()`. Therefore it should be inlined, if there's no justification for it to be a separate method.

src/apscheduler/_schedulers/async_.py
```diff
@@ -529,6 +532,64 @@ async def remove_schedule(self, id: str) -> None:
...
+            schedule.next_fire_time is not None
+            and schedule.next_fire_time >= resume_from
+        ):
+            return schedule.next_fire_time
+        try:
+            while (next_fire_time := schedule.trigger.next()) < resume_from:
+                pass  # Advance `next_fire_time` until its at or past `resume_from`
+        except TypeError:  # The trigger is exhausted
+            return None
+        return next_fire_time
```
agronholm, line 562: Add a blank line after each control block to visually distinguish it from others.

agronholm (changes requested):
Getting there now code-wise. I'll also require a changelog entry (`docs/versionhistory.rst`).

What about the documentation? I think this feature should be documented. Do you think you'd be up for that, or do you want me to do it?

src/apscheduler/_schedulers/async_.py
agronholm: You're right, I've been using `id` for other such parameter names. It's okay to revert.

src/apscheduler/_schedulers/async_.py
agronholm: Also, I don't like catching `TypeError` here, as it has the potential to mask actual bugs. I'd prefer an explicit check against a `None` return value, even if it adds a line or two of code.

WillDaSilva:
@agronholm Thanks for the review. I've addressed your comments related to the code, added an entry in the version history, and added a section about pausing and unpausing schedules to the docs. Let me know if you would like any changes to them, or if you would rather just update them yourself in a follow-up if my writing style in the docs isn't to your liking.

agronholm (approved):
Thanks, I couldn't have written the docs better myself!

PR #905, opened 2024-05-10
fix RedisJobStore .add_job() .update_job() param error

tomnkey:
redis zadd function score pairs should be specified in two ways:
  - As *args, in the form of: name1, score1, name2, score2, ...
  - or as **kwargs, in the form of: name1=score1, name2=score2, ...

### DIFF
--- a/apscheduler/jobstores/redis.py
+++ b/apscheduler/jobstores/redis.py
@@ -82,7 +82,7 @@ class RedisJobStore(BaseJobStore):
                                                           self.pickle_protocol))
             if job.next_run_time:
                 pipe.zadd(self.run_times_key,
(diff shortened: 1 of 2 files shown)


tomnkey:
fix RedisJobStore .add_job() .update_job() param error

agronholm:
Where did you get this idea?

agronholm:
As you can see in the [documentation](https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.zadd), the second argument should be a mapping of keys to values, which is what APScheduler is passing it. As further proof, see the failing tests here.

What prompted this PR in the first place?

PR #907, opened 2024-05-12
refactor: Create `AsyncExitStack` using `attrs` factory

injust:
A bit nicer to have `attrs` create the `AsyncExitStack`, instead of assigning `self._exit_stack = exit_stack.pop_all()` from within the `with` block.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -30,6 +30,8 @@ APScheduler, see the :doc:`migration section <migration>`.
   (PR by MohammadAmin Vahedinia)
 - Fixed the shutdown procedure of the Redis event broker
 - Fixed ``SQLAlchemyDataStore`` not respecting custom schema name when creating enums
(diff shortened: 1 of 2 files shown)


injust:
Not sure why the tests are failing, doesn't seem to be relevant to my change.

agronholm:
You're changing the semantics for the worse. The reason I'm doing `async with AsyncExitStack():` in `__aenter__()` is to ensure that the stack gets rewound if there's an unhandled exception during `__aenter__()`. This would no longer happen if I were to merge this PR.

injust:
> You're changing the semantics for the worse. The reason I'm doing `async with AsyncExitStack():` in `__aenter__()` is to ensure that the stack gets rewound if there's an unhandled exception during `__aenter__()`. This would no longer happen if I were to merge this PR.

Thanks for the explanation; I learned a thing today. Didn't think hard enough about exception handling _within_ `__aenter__()`.

PR #916, opened 2024-05-16
Removed backtick

davidbrochart:
## Changes

Removes an unwanted backtick in documentation.


agronholm:
Weird, the pygrep-hooks pre-commit hook should've caught this. Thanks.

PR #948, opened 2024-07-29
Fixed acquire_jobs method in MongoDB DataStore.

HK-Mattew:
## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ docs/_build/
 build/
 virtualenv/
 venv*/
+env*/
 example.sqlite
--- a/src/apscheduler/datastores/mongodb.py
+++ b/src/apscheduler/datastores/mongodb.py
@@ -686,11 +686,7 @@ class MongoDBDataStore(BaseExternalDataStore):
                         continue
 
                     # Skip and un-acquire the job if no more slots are available
-                    task_slots_left = task_job_slots_left.get(job.task_id, float("inf"))
-                    if (
-                        not task_slots_left
-                        or running_job_count_increments[job.task_id] == task_slots_left
-                    ):
+                    if not task_job_slots_left.get(job.task_id, float("inf")):
                         self._logger.debug(
                             "Skipping job %s because task %r has the maximum "
                             "number of %d jobs already running",
--- a/src/apscheduler/datastores/mongodb.py
+++ b/src/apscheduler/datastores/mongodb.py
@@ -628,7 +628,9 @@ class MongoDBDataStore(BaseExternalDataStore):
                     )
                     async for doc in cursor:
                         task_max_running_jobs = doc["max_running_jobs"]
-                        task_job_slots_left[doc["_id"]] = (doc["max_running_jobs"] - doc["running_jobs"])
+                        task_job_slots_left[doc["_id"]] = (
+                            doc["max_running_jobs"] - doc["running_jobs"]
+                        )
 
                 acquired_jobs: list[Job] = []
                 skipped_job_ids: list[UUID] = []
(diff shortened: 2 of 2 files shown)


src/apscheduler/datastores/mongodb.py
```diff
@@ -684,10 +686,16 @@ async def acquire_jobs(
                         continue
 
                     # Skip and un-acquire the job if no more slots are available
-                    if not task_job_slots_left.get(job.task_id, float("inf")):
+                    task_slots_left = task_job_slots_left.get(job.task_id, float("inf"))
+                    if (
+                        not task_slots_left
+                        or running_job_count_increments[job.task_id] == task_slots_left
+                    ):
```
agronholm, line 693: I think this would be the correct fix?
```suggestion
                    if task_job_slots_left.get(job.task_id, float("inf")):
```

.gitignore
```diff
@@ -14,4 +14,5 @@ docs/_build/
 build/
 virtualenv/
 venv*/
+env*/
```
agronholm, line 17: Let's not.
```suggestion
```

agronholm (changes requested):
Left some comments, and you still need to add a test that would fail without the fix, and a changelog note.

src/apscheduler/datastores/mongodb.py
agronholm: The problem with my current approach with MongoDB is the lack of atomicity. Other schedulers might change the task's running jobs count while this scheduler is running `acquire_jobs()`, as there is no locking involved. I think a better way to do this is to run a conditional update that only updates the running job count if it's lower than the maximum job count. My initial attempt at this failed because `find_and_update()`, for whatever bizarre reason, doesn't support referencing existing fields in the filter. But the next better way would be to use the previously fetched maximum running job counts, as those rarely change.

agronholm: > From what I understand, your change to
> 
> ```python
> if task_job_slots_left.get(job.task_id, float("inf")):
> ```
> 
> Would cause all tasks with a positive value, such as: 1, 2, 3 etc..., to be skipped.
> 
> Am I right or not?

You're right – I was admittely tired when writing that reply. But I think that part of my original logic was correct (`if not task_job_slots_left.get(job.task_id, float("inf")):`). What was wrong was my calculation of free slots (`task_job_slots_left[doc["_id"]] = doc["max_running_jobs"]`), as I should've deducted the number of slots in use from the max slots.

agronholm:
> @agronholm
> 
> You mentioned the problem of not having atomicity when incrementing running_jobs in the task. So I tried some possibilities to increment running_jobs atomically in mongodb. And I got a result that seems good to me.
> 
> See my gist, if you are interested: https://gist.github.com/HK-Mattew/098f714191b348ba6f424ea7add83a1a
> 
> This way it is possible to do the update operation atomically.

I tried something like this, only to find out that you can't use `$expr` in an update filter. Did you actually try that query?

Another criticism is that if only one slot is available and the initial query picks two jobs for that task, both would be dropped here.

agronholm:
> > Another criticism is that if only one slot is available and the initial query picks two jobs for that task, both would be dropped here.
> 
> Indeed. However, I see that it is the best option at the moment. At least to avoid increasing running_jobs incorrectly and adding more jobs.

What I tried myself was doing the increments one by one, and then if the update count is 0, then we can conclude that the task didn't have any more room, yes?

agronholm:
Alright, sounds like we're on the same page now. I'll experiment with your query tomorrow, or you can do it if you like in this PR.

agronholm:
This is what I ended up with (passes all the existing tests):
```py
                    # Try to increment the task's running jobs count
                    update_task_result = await to_thread.run_sync(
                        lambda: self._tasks.update_one(
                            {
                                "_id": job.task_id,
                                "$or": [
                                    {"max_running_jobs": None},
                                    {
                                        "$expr": {
                                            "$gt": [
                                                "$max_running_jobs",
                                                "$running_jobs",
                                            ]
                                        }
                                    },
                                ],
                            },
                            {"$inc": {"running_jobs": 1}},
                            session=session,
                        )
                    )
                    if not update_task_result.matched_count:
                        self._logger.debug(
                            "Skipping job %s because task %r has the maximum number of "
                            "jobs already running",
                            job.id,
                        )
                        skipped_job_ids.append(job.id)
                        continue
```

agronholm:
Right, the `job.task_id` argument was missing. I've added that now. I'll also add a regression test that fails with the unmodified code.

agronholm:
Alright, it's fixed now. Thanks!

PR #965, opened 2024-09-18
Add create shcmea for sqlalchemy datastore

zhu0629:
## Changes
Add create shcmea for sqlalchemy datastore

Fixes #. 

## Checklist

- [  ] You've added tests (in `tests/`) added which would fail without your patch
- [  ] You've updated the documentation (in `docs/`, in case of behavior changes or new
features)
- [ x ] You've added a new changelog entry (in `docs/versionhistory.rst`).

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -60,6 +60,8 @@ APScheduler, see the :doc:`migration section <migration>`.
   useful and reasonable
 - Fixed race condition in ``MongoDBDataStore`` that allowed multiple schedulers to
   acquire the same schedules at once
+- Changed ``SQLAlchemyDataStore`` to automatically create the explicitly specified
+  schema if it's missing
 
 **4.0.0a5**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -61,7 +61,7 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed race condition in ``MongoDBDataStore`` that allowed multiple schedulers to
   acquire the same schedules at once
 - Changed ``SQLAlchemyDataStore`` to automatically create the explicitly specified
-  schema if it's missing
+  schema if it's missing (PR by @zhu0629)
 
 **4.0.0a5**
 
--- a/src/apscheduler/datastores/sqlalchemy.py
+++ b/src/apscheduler/datastores/sqlalchemy.py
@@ -232,12 +232,6 @@ class SQLAlchemyDataStore(BaseExternalDataStore):
 
             yield conn
 
-    async def _create_schema(self, conn: Connection | AsyncConnection) -> None:
-        if not self.schema:
-            return
-        t = CreateSchema(name=self.schema, if_not_exists=True)
-        await self._execute(conn, t)
-
     async def _create_metadata(self, conn: Connection | AsyncConnection) -> None:
         if isinstance(conn, AsyncConnection):
             await conn.run_sync(self._metadata.create_all)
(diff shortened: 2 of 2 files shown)


src/apscheduler/datastores/sqlalchemy.py
```diff
@@ -231,6 +232,13 @@ async def _begin_transaction(
 
             yield conn
 
+    async def _create_schema(self, conn: Connection | AsyncConnection) -> None:
+        if not self.schema:
+            return
+
+        t = text(f"CREATE SCHEMA IF NOT EXISTS {self.schema};")
```
agronholm, line 239: If I am to merge this PR, this needs to use SQLAlchemy's own DDL: https://docs.sqlalchemy.org/en/20/core/ddl.html#sqlalchemy.schema.CreateSchema

agronholm:
You checked the boxes for having updated the documentation and added a changelog. Can you point out to me where you've done this?

agronholm:
Please don't add unrelated changes to this PR.

agronholm:
I will make the necessary updates in `master` and you can then merge them here. I'll let you know when it's done.

agronholm:
I've updated `master` now, and you can rebase your work against that.

docs/versionhistory.rst
```diff
@@ -60,6 +60,7 @@ APScheduler, see the :doc:`migration section <migration>`.
   useful and reasonable
 - Fixed race condition in ``MongoDBDataStore`` that allowed multiple schedulers to
   acquire the same schedules at once
+- Add create `schema` for sqlalchemy datastore
```
agronholm, line 63:
```suggestion
- Changed ``SQLAlchemyDataStore`` to automatically create the explicitly specified schema if it's missing
```

src/apscheduler/datastores/sqlalchemy.py
```diff
@@ -231,6 +232,12 @@ async def _begin_transaction(
 
             yield conn
 
+    async def _create_schema(self, conn: Connection | AsyncConnection) -> None:
+        if not self.schema:
+            return
+        t = schema.CreateSchema(name=self.schema, if_not_exists=True)
```
agronholm, line 238: I'd prefer that you imported this like `from sqlalchemy.schema import CreateSchema`.

src/apscheduler/datastores/sqlalchemy.py
```diff
@@ -389,6 +396,7 @@ async def start(
         async for attempt in self._retry():
             with attempt:
                 async with self._begin_transaction() as conn:
+                    await self._create_schema(conn)
```
agronholm, line 399: Is there a reason for this code to be in its own method? Is there a use case for making it overrideable by the user?

agronholm (changes requested):

agronholm:
Please merge from master because your force pushes are now undoing the updates I made.

src/apscheduler/datastores/sqlalchemy.py
agronholm: You didn't answer my question yet.

agronholm: I just need to resolve this question, and then I'm willing to merge this PR. If indeed you want to make schema creation overrideable by users, then the condition should be here, and the method should just create the schema unconditionally. But I need to understand why you made it into a separate method here.

zhu0629: I will add a param for this tmr.

agronholm:
Could you please stop force-pushing your changes? It's causing some previous updates from `master` from being undone. There's no reason to do that, as I will squash the changes when merging anyway.

src/apscheduler/datastores/sqlalchemy.py
agronholm: What parameter? I wasn't asking about any parameters. And don't resolve the conversation when my question remains unanswered.

zhu0629: I think a separate method is easier to maintain.

agronholm: No, it's just more code. The only plausible reason for keeping it in a separate method is to allow for users to override the behavior, but I don't currently see any reason to let users do that. Additionally, that would increase the API footprint which I don't want to do unless there is a good reason.

zhu0629: Can you modify it according to your suggestion？ thank you

agronholm: I've made the changes I wanted. I'll come back a while later to check on this and merge.

agronholm (approved):

agronholm:
Thanks!

PR #981, opened 2024-10-28
Fix a daylight savings time issue in `CronTrigger`

### LINKED ISSUE #529: DST transitions cause APScheduler to incorrectly calculate next trigger times
**Description**
DST transitions cause APScheduler to incorrectly calculate next trigger times.

Given:
a cron trigger of month='', day='', day_of_week='0-4', hour='0', minute='0', 
a last trigger time of None
a current time of pd.Timestamp(2021, 3, 12, 0, 0,0,1,0,pytz.timezone('US/Central'))

APScheduler returns '2021-03-16T00:00:00-05:00' for the cron_trigger's get_next_fire_time() call.

*NOTE: there is a DST transition CST -> CDT at 2AM on 14 March.*
```
import pytz
import pandas as pd

cron_trigger = CronTrigger.from_crontab('0 0 * * 0-4', pytz.timezone('US/Central'))
now = pd.Timestamp(2021, 3, 12, 0, 0,0,1,0,pytz.timezone('US/Central'))
print(cron_trigger.get_next_fire_time(None,now).isoformat())
```

**Expected behavior**
To return '2021-03-15T00:00:00-05:00' not '2021-03-16T00:00:00-05:00'

**Additional context**
The APScheduler library determines the next fire time by going through each part of the expression and finding the next viable value iteratively until all parts of the expression are satisfied. When the day is incremented, 24 hours are added. This causes the time to go forward 1 day and 1 hour as DST causes a day to only contain 23 hours.

It seems that the _increment_field_value function needs to be corrected to accommodate DST changes
*NOTE: the same behavior doesn't happen on CDT->CST transitions*

### LINKED ISSUE #980: Bi-hourly `CronTrigger` runs into infinite loop at daylight savings time boundary
### Things to check first

### Version

4.0.0a5

### What happened?

When using a bi-hourly `CronTrigger` as given below, starting at a point before the daylight savings time transition, the trigger fails to compute the sequence of next trigger times. It always returns the same static time+date.

The output of the example code given below is:

```
Trigger starts at 2024-10-27 01:00:00+01:00
Next trigger time is 2024-10-27 01:00:00+01:00
Next trigger time is 2024-10-27 01:00:00+00:00
Next trigger time is 2024-10-27 01:00:00+00:00
Next trigger time is 2024-10-27 01:00:00+00:00
Trigger is in future after daylight saving time transition: False
```

Together with [the aync scheduler logic here](https://github.com/agronholm/apscheduler/blob/a79e0275ba13bd733d974db0db67728c632781bc/src/apscheduler/_schedulers/async_.py#L981), this leads to an infinite loop, since the computed next fire time will never become bigger than `now`:

```
                        while True:
                            try:
                                fire_time = calculate_next()
                            except Exception:
                                ...
                                break

                            # Stop if the calculated fire time is in the future
                            if fire_time is None or fire_time > now:
                                next_fire_time = fire_time
                                break
```

The root cause is that the local time `1:00` really exists twice in the given timezone: Once before (`01:00:00+01:00`) and once after the daylight savings time transition (`01:00:00+00:00`).

The trigger stays stuck on the occurrence _after_ the transition.

Run this example

```
import datetime
import zoneinfo

start_time = datetime.datetime(
    2024, 10, 27, 1, 0, 0, 0, tzinfo=zoneinfo.ZoneInfo(key="Europe/Lisbon")
)
print(f"Trigger starts at {start_time}")
trigger = CronTrigger(
    year="*",
    month="*",
    day="*",
    week="*",
    day_of_week="*",
    hour="1,3,5,7,9,11,13,15,17,19,21,23",
    minute="0",
    second="0",
    start_time=start_time,
    timezone="Europe/Lisbon",
)

print(f"Next trigger time is {trigger.next()}")
print(f"Next trigger time is {trigger.next()}")
print(f"Next trigger time is {trigger.next()}")
print(f"Next trigger time is {trigger.next()}")
now_time_after_dst_change = datetime.datetime(
    2024, 10, 27, 14, 0, 0, 0, tzinfo=zoneinfo.ZoneInfo(key="UTC")
)
print(f"Trigger is in future after daylight saving time transition: {trigger.next() > now_time_after_dst_change}")
```

hlobit:
## Changes

Fixes #529.

### DIFF
--- a/src/apscheduler/triggers/cron/__init__.py
+++ b/src/apscheduler/triggers/cron/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Sequence
-from datetime import datetime, timedelta, tzinfo
+from datetime import datetime, tzinfo
 from typing import Any, ClassVar
 
 import attrs
@@ -207,16 +207,17 @@ class CronTrigger(Trigger):
                 else:
                     values[field.name] = new_value
 
-        return datetime(**values, tzinfo=self.timezone)
+        return datetime(**values, tzinfo=self.timezone).replace(fold=dateval.fold)
 
     def next(self) -> datetime | None:
         if self._last_fire_time:
-            start_time = self._last_fire_time + timedelta(microseconds=1)
+            next_time = datetime.fromtimestamp(
+                self._last_fire_time.timestamp() + 1, self.timezone
+            )
         else:
-            start_time = self.start_time
+            next_time = self.start_time
 
         fieldnum = 0
-        next_time = datetime_ceil(start_time).astimezone(self.timezone)
         while 0 <= fieldnum < len(self._fields):
             field = self._fields[fieldnum]
             curr_value = field.get_value(next_time)
@@ -276,11 +277,3 @@ class CronTrigger(Trigger):
 
         fields.append(f"timezone={timezone_repr(self.timezone)!r}")
         return f'CronTrigger({", ".join(fields)})'
-
-
-def datetime_ceil(dateval: datetime) -> datetime:
-    """Round the given datetime object upwards."""
-    if dateval.microsecond > 0:
-        return dateval + timedelta(seconds=1, microseconds=-dateval.microsecond)
-
-    return dateval
--- a/src/apscheduler/triggers/cron/__init__.py
+++ b/src/apscheduler/triggers/cron/__init__.py
@@ -207,7 +207,7 @@ class CronTrigger(Trigger):
                 else:
                     values[field.name] = new_value
 
-        return datetime(**values, tzinfo=self.timezone).replace(fold=dateval.fold)
+        return datetime(**values, tzinfo=self.timezone, fold=dateval.fold)
 
     def next(self) -> datetime | None:
         if self._last_fire_time:
--- a/tests/triggers/test_cron.py
+++ b/tests/triggers/test_cron.py
@@ -401,7 +401,7 @@ def test_dst_change(
 
 
 @pytest.mark.parametrize(
-    "hour, start_time, correct_next_dates",
+    "minute, start_time, correct_next_dates",
     [
         (
             0,
@@ -425,12 +425,12 @@ def test_dst_change(
     ids=["dst_change_0", "dst_change_1"],
 )
 def test_dst_change2(
-    hour,
+    minute,
     start_time,
     correct_next_dates,
     timezone,
 ):
-    trigger = CronTrigger(hour=hour, timezone=timezone)
+    trigger = CronTrigger(minute=minute, timezone=timezone)
     trigger.start_time = start_time.replace(tzinfo=timezone)
     for correct_next_date, fold in correct_next_dates:
         correct_next_date = correct_next_date.replace(tzinfo=timezone, fold=fold)
--- a/tests/triggers/test_cron.py
+++ b/tests/triggers/test_cron.py
@@ -431,10 +431,10 @@ def test_dst_change2(
     timezone,
 ):
     trigger = CronTrigger.from_crontab(cron_expression, timezone=timezone)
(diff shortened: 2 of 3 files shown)


agronholm:
How did you get the test to pass for you locally? Are the new tests sensitive to the local time zone?

agronholm:
Either way, I'll take a look at this tomorrow.

agronholm:
Oh, and I'll need a new changelog entry for this too (don't delete the pull request template, as it contains a check list of things for the PR to be accepted).

agronholm:
Oops, it's been 2 weeks already! I got distracted by several other projects I'm working on. Anyways, the workings of the cron trigger are not fresh in my memory, so would you mind explaining these changes? They look fine, but I'd like to understand why these specific changes were necessary.

hlobit:
No problem !
I think what is key in this change is the replacement of a statement like

```python
self._last_fire_time + timedelta(seconds=1)
```

by

```python
datetime.fromtimestamp(
    self._last_fire_time.timestamp() + 1, self.timezone
)
```

For any reason, the first one does not take DST into account.
https://docs.python.org/3/library/datetime.html#datetime.datetime.fold
> Note that no time zone adjustments are done even if the input is an aware object.

The second one does the trick because it is doing operations using the `timestamp()`, which handles the `fold` attribute.
https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp
> The [timestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp) method uses the [fold](https://docs.python.org/3/library/datetime.html#datetime.datetime.fold) attribute to disambiguate the times during a repeated interval.

The other change is to make sure the value of `fold` attribute is not lost while returning a new aware `datetime` object from `_set_field_value()`.

This made my use cases pass, see the tests.

And thanks for the great apscheduler :heart:

tests/triggers/test_cron.py
```diff
@@ -400,6 +400,44 @@ def test_dst_change(
...
+    start_time,
+    correct_next_dates,
+    timezone,
+):
+    trigger = CronTrigger.from_crontab(cron_expression, timezone=timezone)
+    trigger.start_time = start_time.replace(tzinfo=timezone)
+    for correct_next_date, fold in correct_next_dates:
+        next_date = trigger.next()
+        assert next_date == correct_next_date.replace(tzinfo=timezone)
+        assert next_date.fold == fold
```
agronholm, line 438: This would make me feel better about the correctness.
```suggestion
        assert next_date == correct_next_date.replace(tzinfo=timezone, fold=fold)
        assert str(next_date) == str(correct_next_date)
```

tests/triggers/test_cron.py
```diff
@@ -400,6 +400,44 @@ def test_dst_change(
     )
 
 
+@pytest.mark.parametrize(
+    "cron_expression, start_time, correct_next_dates",
+    [
+        (
+            "0 * * * *",
```
agronholm, line 407: Why bother with `from_crontab()`? This could just be the `minute` field.

src/apscheduler/triggers/cron/__init__.py
```diff
@@ -207,16 +207,17 @@ def _set_field_value(
                 else:
                     values[field.name] = new_value
 
-        return datetime(**values, tzinfo=self.timezone)
+        return datetime(**values, tzinfo=self.timezone).replace(fold=dateval.fold)
```
agronholm, line 210:
```suggestion
        return datetime(**values, tzinfo=self.timezone, fold=dateval.fold)
```

agronholm (changes requested):
Nice work, but I'd like a few tweaks before I merge.

agronholm (approved):

agronholm:
Thanks! This was a big help.

PR #982, opened 2024-10-29
Fix tests

HK-Mattew:
Removed failed test that checked if the Scheduler was running on uwsgi with threads disabled. Now in new versions of uwsgi threads are enabled by default.

Reference to uwsgi change: https://github.com/unbit/uwsgi/commit/896ee575d0743e78237c7220007f4bbeaa8cedf8

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ test = [
     "pytest-mock",
     "time-machine >= 2.13.0; python_implementation == 'CPython'",
     """\
-    uwsgi; python_implementation == 'CPython' and platform_system == 'Linux'\
+    uwsgi <= 2.0.26; python_implementation == 'CPython' and platform_system == 'Linux'\
     and python_version < '3.13'\
     """,
 ]


HK-Mattew:
After I opened the PR, I thought about the possible "fix". Instead of removing test, I could just skip testing for newer versions of Uwsgi.

I didn't think about this before, but if anyone prefers I can try to do it later.

agronholm:
Perhaps you could just disable threads explicitly? The scheduler should be allowed to run if uwsgi has threads enabled, and fail if it doesn't, regardless of version.

HK-Mattew:
> Perhaps you could just disable threads explicitly? The scheduler should be allowed to run if uwsgi has threads enabled, and fail if it doesn't, regardless of version.

I had looked for an option in uwsgi to disable threads. And I believe there is no such option or I did not find it

HK-Mattew:
> Perhaps you could just disable threads explicitly? The scheduler should be allowed to run if uwsgi has threads enabled, and fail if it doesn't, regardless of version.

agronholm:
In that case, the `uwsgi` test dependency should have an upper bound where it avoids any version with thread enablement by default.

HK-Mattew:
Actually, in new versions of uwsgi I don't know yet when the value of has_threads(https://github.com/agronholm/apscheduler/blob/a79e0275ba13bd733d974db0db67728c632781bc/src/apscheduler/_schedulers/sync.py#L395) can be negative.

HK-Mattew:
Sorry, I ended up closing the PR by accident 🤦‍♂️

HK-Mattew:
> In that case, the `uwsgi` test dependency should have an upper bound where it avoids any version with thread enablement by default.

Do you mean, skip this specific test for new versions of uwsgi?

agronholm:
No, I mean install an old version of uwsgi that still disables threading by default.

HK-Mattew:
> No, I mean install an old version of uwsgi that still disables threading by default.

Done!

agronholm:
Thanks!

PR #990, opened 2024-11-18
Fix test_executor failing tests when using python 3.14

majamassarini:
Fedora is introducing python 3.14 and some tests are failing: https://bugzilla.redhat.com/show_bug.cgi?id=2325181

Should fix https://bugzilla.redhat.com/show_bug.cgi?id=2325181

## Changes

As far as I can understand it, this are the issues:
 - The `ProcessPoolExecutor` is reusing the same worker to execute different tasks and for this reason the comparison between the tasks id is failing.
- Some tests end before their task completion.
- I have to adjust the `misfire_grace_time` to make some tasks execute (don't know why this wasn't needed before).

### DIFF
--- a/tests/test_executors.py
+++ b/tests/test_executors.py
@@ -36,11 +36,11 @@ def executor(request, mock_scheduler):
         executor_ = ThreadPoolExecutor()
     else:
         from apscheduler.executors.pool import ProcessPoolExecutor
-        executor_ = ProcessPoolExecutor()
+        executor_ = ProcessPoolExecutor(4, {"max_tasks_per_child": 1})
 
     executor_.start(mock_scheduler, 'dummy')
     yield executor_
-    executor_.shutdown()
+    executor_.shutdown(True)
 
 
 def wait_event():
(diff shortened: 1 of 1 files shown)


agronholm:
I've pushed a ton of updates to the 3.x branch. Would you mind merging them to this branch?

agronholm:
The root cause seems to have been that in 3.14, `ProcessPoolExecutor` uses the `spawn` method instead of `fork`, so the time freeze fixture could not affect the subprocess. I've adjusted the tests accordingly in the updated code, in addition to forcing the use of `spawn` instead of `fork` in the executor.

agronholm:
I just ran the tests on 3.14.0a1 and the executor tests all passed after today's code changes, so doesn't look like this PR is necessary. I'm going to close it, but let me know if you find otherwise.

majamassarini:
Could you release a new apscheduler 3.10.5 (or whatever version you think is right) so I can package it downstream with your fixes? 

Thank you.

agronholm:
I'll make a release soon. I just have a few more changes to add to it.

Issue #952, opened 2024-08-09
App crash


mmmcorpsvit:
### Things to check first

- [X] I have checked that my issue does not already have a solution in the [FAQ](https://apscheduler.readthedocs.io/en/master/faq.html)

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Version

4.0.0a5

### What happened?

work app

### How can we reproduce the bug?

```python
from asyncio import run, sleep
from datetime import datetime

from apscheduler import AsyncScheduler
from apscheduler.triggers.interval import IntervalTrigger

async def tick():
    await sleep(1000)
    print("Hello, the time is", datetime.now())


async def main():
    async with AsyncScheduler() as scheduler:
        # await scheduler.add_schedule(parser_copart_collect_csv_run2, IntervalTrigger(seconds=60))
        await scheduler.add_schedule(tick, IntervalTrigger(seconds=1))
        await scheduler.run_until_stopped()


run(main())
```

after some time app crash (windows 11, python 3.11.9):

mmmcorpsvit:
```txt
C:\Users\user\AppData\Local\pypoetry\Cache\virtualenvs\cars-pRqyGtAI-py3.11\Scripts\python.exe D:\GoogleDrive\projects\cars-backend\worker.py 
INFO:apscheduler._schedulers.async_:Added new schedule (task='__main__:tick', trigger=IntervalTrigger(seconds=1, start_time='2024-08-10 01:49:19.465905+03:00')); next run time at 2024-08-10 01:49:19.465905+03:00
INFO:apscheduler._schedulers.async_:Cleaned up expired job results and finished schedules
INFO:apscheduler._schedulers.async_:Scheduler started
INFO:apscheduler._schedulers.async_:Cleaned up expired job results and finished schedules
Hello, the time is 2024-08-10 02:05:59.468589
INFO:apscheduler._schedulers.async_:Job 2474e627-e5ed-4994-a506-1cec5575e37e completed successfully
INFO:apscheduler._schedulers.async_:Job 0be9b4c3-101d-4382-bd2d-8dedcf6f1603 completed successfully
Hello, the time is 2024-08-10 02:06:00.479031
Hello, the time is 2024-08-10 02:06:01.476233
INFO:apscheduler._schedulers.async_:Job e38794ac-433d-4f89-845e-715ad72a6fd6 completed successfully

Hello, the time is 2024-08-10 02:06:12.497892
[...]

PR #991, opened 2024-11-19
Avoid crash when removing job which might not exist in async_.py

neeraj9:
I found a scenario once though harder to reproduce that scheduler crashes when a job runs longer than the schedule which has an interval of say 1 minute. I believe the duration is lesser important, but it is possible that the job runs longer than 1 minute and when the time comes to remove it then it do not exist and calling python `set.remove(id)` crashes the scheduler which should not be so. Hence protecting it.

## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -1216,4 +1216,4 @@ class AsyncScheduler:
             finally:
                 current_job.reset(token)
         finally:
-            self._running_jobs.remove(job)
+            self._running_jobs.discard(job)


HK-Mattew:
I don't think this is a fix. You're just skipping the problem.

agronholm:
> I don't think this is a fix. You're just skipping the problem.

I concur. @neeraj9 I don't remember if I've fixed this in master yet – have you verified that it still occurs there?

neeraj9:
In principle I agree that this does hide the issue, so it should be logged and fixed appropriately. Having said that its worse to crash because restarting scheduler in a fastapi setup is non-trivial (at least for me when I have not spent much time with apscheduler).

neeraj9:
> > I don't think this is a fix. You're just skipping the problem.
> 
> I concur. @neeraj9 I don't remember if I've fixed this in master yet – have you verified that it still occurs there?

I have not tested with master directly, but got an error with version 4.0.0a5 available from pip.

HK-Mattew:
I just checked and the error no longer occurs for me on the master branch. I even commented on the issue: https://github.com/agronholm/apscheduler/issues/952#issuecomment-2485617250

agronholm:
Yup, closing as redundant.

Issue #953, opened 2024-08-10
Add .get_task method to Schedulers - APS v4


mattewid:
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

Hello,

My suggestion is to add the .get_task(task_id=...) method to the Schedulers.

### Use case

I found myself in a situation where I needed to pass the Task instance directly to the .add_job method to get an existing task configuration.

I could use the method to get all tasks with .get_tasks. But I would have to filter this list every time to get a single specific task. I don't think this would be a very interesting approach in my use case and I believe my suggestion will be useful to others as well.

agronholm:
I'll consider this, but I'm curious as to why you would need to pass a `Task` instance to `add_job()`. Could you explain that?

agronholm:
Passing the task ID to `add_job()` should never overwrite any task configuration. Can you give me a reproducible example where you demonstrate such behavior?

agronholm:
Ok, I understand the problem now, and it's a design issue. I'll have to refactor the `add_task()` data store method.

agronholm:
I'm making some progress once in a while, but it seems that every time I fix something, I uncover another problem. The rabbit hole is deep :frowning_face: 
I'll get it done Soon(tm). But I have people in other projects constantly asking for updates, not just APScheduler...

agronholm:
The hard work on AnyIO's next release is done, so I can focus on this now. Getting incremental updates to task configuration is the crux of the problem here. I'm still working on a solution to that.

agronholm:
Sorry for the delay. I'm having a bit of trouble fixing `AsyncScheduler.configure_task()` to work with the data stores in a sane way, so I'm currently experimenting with different ways of implementing full and partial task updates. This might take some time, unfortunately.

Issue #465, opened 2020-09-29
APScheduler 4.0 progress tracking


agronholm:
I'm opening this issue as an easy way to interested parties to track development progress of the next major APScheduler release (v4.0).

## Terminology changes in v4.0

The old term of "Job", as it was, is gone, replaced by the following concepts which are closer to the terminology used by Celery:

- Task definition: a uniquely named callable coupled with configuration like maximum number of instances, misfire grace time etc.
- Schedule: binds a trigger with a task definition
- Job: queued work item for an executor (binds to a task definition, and optionally a schedule)

Notice that the terminology may still change before the final release!

## Planned major changes

v4.0 is a ground-up redesign that aims to fix all the long-standing flaws found in APScheduler over the years.

Checked boxes are changes that have already been implemented.

- [X] Async-first design, with support for asyncio, curio and trio (via [AnyIO](https://github.com/agronholm/anyio))
- [X] Static typing friendly ([PEP 561](https://www.python.org/dev/peps/pep-0561/))
- [X] Support for serializers other than pickle
- [X] Broader time zone support, including zoneinfo time zones ([PEP 615](https://www.python.org/dev/peps/pep-0615/))
- [X] Drop support for Python < 3.7
- [X] [Calendar interval trigger](https://github.com/agronholm/apscheduler/blob/master/apscheduler/triggers/calendarinterval.py)
- [X] Stateful triggers
- [X] `threshold` value for `AndTrigger` (resolves issues with contained `IntervalTrigger` instances)
- [X] The interval trigger should start right away and not after the first interval (#375)
- [ ] Persistent store sharing among multiple schedulers (arguably the most needed feature ever for APScheduler)
- [ ] Context-local job metadata information
- [ ] Easy launching of tasks immediately without needing a schedule

Potential extra features I would like to have:

- [ ] Support for tags in task definitions, schedules and jobs
- [ ] "threshold" value for `OrTrigger` (#427)

You will notice that I have dropped a number of features from master. Some I may never add back to v4.0, even if requested, but do voice your wishes in this issue (and this issue **only** – I will summarily close such requests in new tickets). Others have been removed only temporarily to give me space for the redesign.

## Features on the chopping block

- Twisted scheduler (may be usable through the async scheduler if AnyIO ever gets Twisted support)
- Tornado scheduler (just use the async scheduler)
- Gevent scheduler (does not play well with the new architecture)

agronholm:
[...]

PR #993, opened 2024-11-19
Fixed an issue with task configuration rewriting in the .add_job method

mattewid:
## Changes

Fixes
- Some changes were made to the .configure_task method to fix an issue mentioned in #953

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -75,6 +75,7 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed the job lease extension task exiting prematurely while the scheduler is starting
   (PR by @JacobHayes)
 - Migrated test and documentation dependencies from extras to dependency groups
+- Fixed task configuration rewrite in .add_job/.configure_task method (PR by @mattewid)
 
 **4.0.0a5**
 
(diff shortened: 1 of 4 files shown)


HK-Mattew:
If these changes are accepted, let me know and I'll add the entry to the changelog.

HK-Mattew:
**Hello** @agronholm 

I saw your comment on https://github.com/agronholm/apscheduler/issues/465#issuecomment-2825642271 about the **.configure_task** method. Wouldn't this PR solve the problems with this method?

agronholm:
As I recall I looked at this before but felt it was an incomplete solution. I'll take another look before I proceed any further.

agronholm:
Yeah, the problem I was looking at was that it's not exactly concurrent safe, but it's an improvement to the existing code nonetheless. Just add the changelog entry (third checkbox in the PR form) and I'll merge.

mattewid:
> Yeah, the problem I was looking at was that it's not exactly concurrent safe, but it's an improvement to the existing code nonetheless. Just add the changelog entry (third checkbox in the PR form) and I'll merge.

Yes, from the beginning I knew that it would only be a solution in cases where there was no competition. But even so, it is already a good correction.

Thanks for the analysis. I will do what is missing soon.

PR #1004, opened 2024-12-21
Add cooldown_period to OrTrigger

### LINKED ISSUE #453: Having the threshold also on OrTrigger?
Hi @agronholm ,

Just curious : you added the threshold parameter onto `AndTrigger` but I'm wondering if you plan or if you could add it to the `OrTrigger`?

In my case I may have different intervals that can something be distant from 0, 1 or 2 seconds and I would like to avoid triggering all of them, but just one when in the same threshold.

Thank you,

HomerusJa:
## Changes

Fixes #453

- Added the `cooldown_period` attribute to the OrTrigger, preventing two triggers triggering within that interval.
- Added a `max_iterations` attribute to the OrTrigger as in the AndTrigger preventing near infinite looping when two triggers are triggering a lot whilst being in the cooldown phase.
- Added tests to confirm the behavior

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/apscheduler/triggers/combining.py
+++ b/src/apscheduler/triggers/combining.py
@@ -125,40 +125,53 @@ class OrTrigger(BaseCombiningTrigger):
 
     cooldown_period: timedelta = attrs.field(converter=as_timedelta, default=0)
     _last_fire_time: datetime | None = attrs.field(default=None, eq=False, init=False)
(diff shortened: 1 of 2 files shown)


HomerusJa:
Should I add something to the docs? I suppose the API documentation is automatically generated, so where else should I mention this feature?

Issue #1013, opened 2025-01-09
Unreachable `None` handling in `as_timezone()` converter


injust:
### Things to check first

- [X] I have checked that my issue does not already have a solution in the [FAQ](https://apscheduler.readthedocs.io/en/master/faq.html)

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Version

844dc05

### What happened?

```python
>>> from apscheduler.triggers.cron import CronTrigger
>>> CronTrigger(timezone=None)
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    CronTrigger(timezone=None)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "<attrs generated init apscheduler.triggers.cron.CronTrigger>", line 25, in __init__
    __attr_validator_timezone(self, __attr_timezone, self.timezone)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/foo/.venv/lib/python3.13/site-packages/attr/validators.py", line 100, in __call__
    raise TypeError(
    ...<4 lines>...
    )
TypeError: ("'timezone' must be <class 'datetime.tzinfo'> (got None that is a <class 'NoneType'>).", Attribute(name='timezone', default=Factory(factory=<function get_localzone at 0x10a3f4720>, takes_self=False), validator=<instance_of validator for type <class 'datetime.tzinfo'>>, repr=True, eq=True, eq_key=None, order=True, order_key=None, hash=None, init=True, metadata=mappingproxy({}), type='tzinfo', converter=<function as_timezone at 0x10a3c3880>, kw_only=True, inherited=False, on_setattr=None, alias='timezone'), <class 'datetime.tzinfo'>, None)
```

### How can we reproduce the bug?

```python
from apscheduler.triggers.cron import CronTrigger

CronTrigger(timezone=None)
```

injust:
https://github.com/agronholm/apscheduler/blob/844dc05364052b94e046bff86e9b101bcd2c36c6/src/apscheduler/_converters.py#L40-L43

It seems like `as_timezone(None)` was _intended_ to return `get_localzone()`, but actually falls through and return `None`.

[...]

PR #1014, opened 2025-01-09
Improve converters and type annotations

injust:
## Changes

- Add proper type annotations to converters, instead of `Any`
- `as_timezone`: Remove unreachable `None` handling (fixes https://github.com/agronholm/apscheduler/issues/1013)
	- It doesn't make sense for `None` to do the same thing as the `"local"` default while being less explicit, so I omitted it from the type annotation and dropped the unreachable code
	- The current runtime behaviour is maintained, but a type checker will now warn you about it
- `CronTrigger.from_crontab`: Swap `timezone` parameter type annotation ordering to `tzinfo | str` for consistency
    - This ordering seems more logical than `str | tzinfo`
- `as_enum`: Use `Enum.__getitem__` instead of `Enum.__members__.__getitem__`

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/src/apscheduler/_converters.py
+++ b/src/apscheduler/_converters.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable
-from datetime import date, datetime, timedelta, timezone
+from datetime import date, datetime, timedelta, timezone, tzinfo
 from typing import Any
 from uuid import UUID
 from zoneinfo import ZoneInfo
@@ -9,14 +9,14 @@ from zoneinfo import ZoneInfo
 from tzlocal import get_localzone
 
 
-def as_int(value: Any) -> Any:
+def as_int(value: int | str) -> int:
     if isinstance(value, str):
         return int(value)
 
     return value
 
 
-def as_aware_datetime(value: Any) -> Any:
+def as_aware_datetime(value: datetime | str) -> datetime:
     if isinstance(value, str):
         # Before Python 3.11, fromisoformat() could not handle the "Z" suffix
         if value.upper().endswith("Z"):
@@ -30,33 +30,30 @@ def as_aware_datetime(value: Any) -> Any:
     return value
 
 
-def as_date(value: Any) -> Any:
+def as_date(value: date | str) -> date:
     if isinstance(value, str):
         return date.fromisoformat(value)
 
     return value
 
 
-def as_timezone(value: Any) -> Any:
+def as_timezone(value: tzinfo | str) -> tzinfo:
     if isinstance(value, str):
-        if value is None or value == "local":
-            return get_localzone()
-
-        return ZoneInfo(value)
+        return get_localzone() if value == "local" else ZoneInfo(value)
     elif value is timezone.utc:
         return ZoneInfo("UTC")
 
     return value
 
 
-def as_uuid(value: Any) -> Any:
+def as_uuid(value: UUID | str) -> UUID:
     if isinstance(value, str):
         return UUID(value)
 
     return value
 
 
-def as_timedelta(value: Any) -> Any:
+def as_timedelta(value: timedelta | int) -> timedelta:
     if isinstance(value, (float, int)):
         return timedelta(seconds=value)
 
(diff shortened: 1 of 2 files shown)


injust:
Hmm, I'm not sure why the docs build is failing when I type `as_enum()` using `enum.EnumMeta`:

```
WARNING: Cannot resolve forward reference in type annotations of "apscheduler.Schedule": name 'EnumMeta' is not defined
WARNING: Cannot resolve forward reference in type annotations of "apscheduler.JobResult": name 'EnumMeta' is not defined
WARNING: Cannot resolve forward reference in type annotations of "apscheduler.AsyncScheduler": name 'EnumMeta' is not defined
```

And I'm not 100% sure how to best type `list_converter`.

I'm going to leave those 2 as is.

injust:
I diffed the [current docs](https://apscheduler.readthedocs.io/en/master/api.html) against the [RTD build for this PR](https://apscheduler--1014.org.readthedocs.build/en/1014/api.html) and the changes look good.

injust:
But somehow I made sphinx-autodoc-typehints unhappy -- https://apscheduler--1014.org.readthedocs.build/en/1014/api.html#apscheduler.triggers.interval.IntervalTrigger looks like this:

image

The types are no longer in code blocks and don't link to the Python docs anymore. No idea what happened.

injust:
I'm also convinced that `as_enum()` isn't necessary -- you should be able to use `Enum.__getitem__` as the converter.

However, this doesn't work for `StrEnum`, which I think is a bug (https://github.com/python/cpython/issues/128659).

injust:
`as_enum()` should be typed like this:

```python
def as_enum(enum_class: type[Enum]) -> Callable[[Enum | str], Enum]:
    def converter(value: Enum | str) -> Enum: ...
```

but it breaks the docs build (both local and on RTD):

```
WARNING: Cannot resolve forward reference in type annotations of "apscheduler.Schedule": name 'Enum' is not defined
WARNING: Cannot resolve forward reference in type annotations of "apscheduler.JobResult": name 'Enum' is not defined
WARNING: Cannot resolve forward reference in type annotations of "apscheduler.AsyncScheduler": name 'Enum' is not defined
```

So I'll undo the type annotations but keep the change from `enum_class.__members__[value]` to `enum_class[value]`, since that's how enum members should be accessed by name (https://docs.python.org/3/howto/enum.html#programmatic-access-to-enumeration-members-and-their-attributes).

agronholm:
The type annotations for converters are not appropriate, as converters are run before validators, so the input value could be literally anything.

injust:
> The type annotations for converters are not appropriate, as converters are run before validators, so the input value could be literally anything.

I'm not sure I follow. Type annotations should describe what's _intended_, so your type checker can catch errors before you get a `TypeError` from the validator at runtime.

injust:
> Sure, but does the type checker know that it should look at the `attrs` converter to determine what types are allowed as input?

Absolutely!

image

image

See https://www.attrs.org/en/stable/init.html#converters:

> If a converter’s first argument has a type annotation, that type will appear in the signature for `__init__`. A converter will override an explicit type annotation or type argument.

agronholm:
Since you've verified that these changes improve the type checking and documentation, I'm merging this. Thanks!

injust:
This PR breaks pyright, but it seems to only happen when APScheduler is used in an attrs field like this:

```python
from apscheduler import AsyncScheduler
from attrs import Factory, define


@define
class Foo:
    scheduler: AsyncScheduler = Factory(AsyncScheduler)
```

```
./foo.py:7:41 - error: Argument of type "type[AsyncScheduler]" cannot be assigned to parameter "factory" of type "() -> _T@Factory" in function "Factory"
  Type "type[AsyncScheduler]" is not assignable to type "() -> AsyncScheduler"
    Type "timedelta | None" is not assignable to type "timedelta | int"
      Type "None" is not assignable to type "timedelta | int"
        "None" is not assignable to "timedelta"
        "None" is not assignable to "int" (reportArgumentType)
```

pyright's behaviour seems to be inconsistent, and I'll open a discussion with them to see why this happens and if it's a bug.

I'll also open a follow-up PR to add `attrs.converters.optional()` to fields that should take `None`.

PR #1017, opened 2025-01-13
Make tick routine async

HelgeKrueger:
This helps to illustrate that scheduler can be used to run an async function.


agronholm:
I have mixed feelings about this. The function doesn't _need_ to be async, and users have generally not had any difficulties figuring out that coroutine functions are okay to use (this is the sole reason for the existence of `AsyncIOScheduler` after all).

PR #1025, opened 2025-02-08
Fix "RuntimeError: cannot schedule new futures after interpreter shutdown"

### LINKED ISSUE #985: RuntimeError: cannot schedule new futures after interpreter shutdown
### Things to check first

### Version

3.10.4

### What happened?

Hello,

I'm trying to create a Python code orchestrator that schedules a function to be called 20 seconds after being triggered. However, every time it runs, the log shows a RuntimeError stating that new tasks cannot be scheduled after the interpreter has shut down.

Interestingly, I used this same code with Python 3.7 without any issues. The only change I've made is updating Python to version 3.10 and APScheduler to version 3.10.4.

Here is the error log:
```
[INFO] 2024-11-09 20:02:49,514 - apscheduler.scheduler - Removed job frontend_305_
[INFO] 2024-11-09 20:03:41,713 - apscheduler.scheduler - Scheduler started
[INFO] 2024-11-09 20:03:55,294 - apscheduler.scheduler - Added job "QueeList.execute_process" to job store "default"
[ERROR] 2024-11-09 20:04:15,297 - apscheduler.scheduler - Error submitting job "QueeList.execute_process (trigger: date[2024-11-09 20:04:15 -03], next run at: 2024-11-09 20:04:15 -03)" to executor "default"
  File "C:\Users\gusta\teste\Em andamento\PyControlAPI\backend\nvenv\lib\site-packages\apscheduler\schedulers\base.py", line 988, in _process_jobs
    executor.submit_job(job, run_times)
  File "C:\Users\gusta\teste\Em andamento\PyControlAPI\backend\nvenv\lib\site-packages\apscheduler\executors\base.py", line 71, in submit_job
    self._do_submit_job(job, run_times)
  File "C:\Users\gusta\teste\Em andamento\PyControlAPI\backend\nvenv\lib\site-packages\apscheduler\executors\pool.py", line 28, in _do_submit_job
    f = self._pool.submit(run_job, job, job._jobstore_alias, run_times, self._logger.name)
  File "C:\Users\gusta\AppData\Local\Programs\Python\Python310\lib\concurrent\futures\thread.py", line 163, in submit
    raise RuntimeError('cannot schedule new futures after '
RuntimeError: cannot schedule new futures after interpreter shutdown
```

```
from logging import basicConfig,  INFO, FileHandler
from apscheduler.schedulers.blocking import BlockingScheduler
from datetime import datetime, timedelta
from typing import Optional
import threading


class QueeList:
    def __init__(self, scheduler=None):
        self.scheduler = scheduler
    
    def execute_process(self, process_value):
        print("hello world!")
        
    def reload_proess_database(self):
        list_values = [("*/5 * * * *", "teste", 10, "teste")]
        for cron_value, process_value, process_id, process_name in list_values:
            self.scheduler.add_job(self.execute_process, trigger="date", run_date=datetime.now() + timedelta(seconds=20), args=[process_value], id=f"frontend_{process_id}_{process_name}")


def main_orchestrator_control(save_logs):
    scheduler = BlockingScheduler()
    list_process = QueeList(scheduler)
    list_process.reload_process_database()
    list_process.scheduler.start()

def control_start_orchestrator():
    try:        
        orchestrator = threading.Thread(target=main_orchestrator_control, name="OrchestratorController", args=(False), daemon=True)
        orchestrator.start()
        
    except Exception as erro:
        print("problema ao executar função.")


if __name__ == "__main__":
    VAR_PATH_LOGS_SYSTEM = "C:/temp/TesteOrchestrator.txt"
    VAR_MESSAGE_LOG_FORMAT = "[%(levelname)s] %(asctime)s - %(name)s - %(message)s"

    basicConfig(level=INFO, format=VAR_MESSAGE_LOG_FORMAT, handlers=[FileHandler(VAR_PATH_LOGS_SYSTEM, "a", "utf-8")])

    orchestrator = control_start_orchestrator() 
```

i-am-darshil:
## Changes

Fixes #985

## Summary
This PR addresses a race condition where a job submission coincides with the executor shutting down, leading to unnecessary RuntimeError logs. While the error is harmless, it clutters logs and can be misleading.

### Root Cause
1. A job is scheduled and is about to be submitted to the executor’s thread/process pool.
2. Before submission completes, the executor begins shutting down.
3. The shutdown process acquires _shutdown_lock and marks the executor as shut down.
4. A job submission is attempted post-shutdown, causing a RuntimeError.
```
  File "C:\Users\gusta\teste\Em andamento\PyControlAPI\backend\nvenv\lib\site-packages\apscheduler\schedulers\base.py", line 988, in _process_jobs
    executor.submit_job(job, run_times)
  File "C:\Users\gusta\teste\Em andamento\PyControlAPI\backend\nvenv\lib\site-packages\apscheduler\executors\base.py", line 71, in submit_job
    self._do_submit_job(job, run_times)
  File "C:\Users\gusta\teste\Em andamento\PyControlAPI\backend\nvenv\lib\site-packages\apscheduler\executors\pool.py", line 28, in _do_submit_job
    f = self._pool.submit(run_job, job, job._jobstore_alias, run_times, self._logger.name)
  File "C:\Users\gusta\AppData\Local\Programs\Python\Python310\lib\concurrent\futures\thread.py", line 163, in submit
    raise RuntimeError('cannot schedule new futures after '
RuntimeError: cannot schedule new futures after interpreter shutdown
````

### Steps to reproduce
https://github.com/agronholm/apscheduler/issues/985#issuecomment-2642421776

## Checklist

- [ ✅] You've added tests (in `tests/`) added which would fail without your patch
features)
- [ ✅] You've added a new changelog entry (in `docs/versionhistory.rst`).

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,6 +8,8 @@ APScheduler, see the :doc:`migration section <migration>`.
 
 - Fixed ``scheduler.shutdown()`` not raising ``SchedulerNotRunning`` (or raising the
   wrong exception) for asynchronous schedulers when the scheduler is in fact not running
+- Fixed "RuntimeError: cannot schedule new futures after interpreter shutdown" (#985
+<https://github.com/agronholm/apscheduler/issues/985>_; PR by @i-am-darshil)
 
 **3.11.0**
 
--- a/src/apscheduler/executors/pool.py
+++ b/src/apscheduler/executors/pool.py
@@ -6,6 +6,12 @@ from concurrent.futures.process import BrokenProcessPool
 from apscheduler.executors.base import BaseExecutor, run_job
 
 
+SHUTDOWN_MESSAGES = {
+    "cannot schedule new futures after shutdown",
+    "cannot schedule new futures after interpreter shutdown",
+}
+
+
 class BasePoolExecutor(BaseExecutor):
     @abstractmethod
     def __init__(self, pool):
@@ -24,10 +30,20 @@ class BasePoolExecutor(BaseExecutor):
             else:
                 self._run_job_success(job.id, f.result())
 
-        f = self._pool.submit(
-            run_job, job, job._jobstore_alias, run_times, self._logger.name
-        )
+        try:
+            f = self._pool.submit(
+                run_job, job, job._jobstore_alias, run_times, self._logger.name
+            )
+        except RuntimeError as e:
+            # There is a possible race condition between shutdown being invoked and the job being submitted to pool.
+            # This helps with not polluting logs with RuntimeError traces.
+            if str(e) in SHUTDOWN_MESSAGES:
+                self._logger.warning(f"Executor has been shut down, skipping job {job.id}: {e}")
+                return
+            else:
+                raise  # Re-raise unexpected RuntimeErrors
         f.add_done_callback(callback)
+
 
     def shutdown(self, wait=True):
         self._pool.shutdown(wait)
--- a/tests/test_executors.py
+++ b/tests/test_executors.py
@@ -2,6 +2,7 @@ import gc
 import os
 import signal
 import time
+import logging
 from asyncio import CancelledError
 from datetime import datetime
 from threading import Event
@@ -271,6 +272,25 @@ async def test_run_coroutine_job(asyncio_scheduler, asyncio_executor, exception)
         assert str(events[0].exception) == "dummy error"
     else:
         assert events[0].retval is True
+
+
+@pytest.mark.parametrize('message,raise_error', [
+    ("cannot schedule new futures after shutdown", False),
+    ("cannot schedule new futures after interpreter shutdown", False),
+    ("random RuntimeError", True),
+])
+def test_pool_shutdown(mock_scheduler, executor, create_job, timezone, caplog, message, raise_error):
+    executor._pool.submit = MagicMock(side_effect=RuntimeError(message))
+    with caplog.at_level(logging.WARNING):
+        try:
+            job = create_job(func=wait_event, max_instances=2, next_run_time=None)
+            run_time = datetime.now(timezone)
+            executor.submit_job(job, [run_time])
+            if raise_error:
+                pytest.fail("Random RuntimeErrors should be raised!")
+        except RuntimeError:
+            if not raise_error:
+                pytest.fail("Shutdown RuntimeErrors should be caught and logged, not raised!")
 
 
 @pytest.mark.parametrize("exception", [False, True])


i-am-darshil:
Hey @agronholm, whenever you get a chance, can you 👀 at this?

agronholm:
To me, it seems like this doesn't address the root cause, but simply sweeps the error message under the carpet. A better approach would be to reorganize the shutdown process in such a way that the scheduler won't even **try** to schedule jobs after shutdown, yes? I don't know how feasible this is in v3.x (it's already handled on v4.x) but I'd like to see an attempt made first.

i-am-darshil:
Hey @agronholm,
I didn't want to make big changes initially since you're already rolling out v4.x. But, agreed with your suggestion. Let me try it out.

Quite curious & hence one question, I see v3.10.4 is quite different branch `v3.x`. However, the only active branch is `3.x`
How are the changes from v3.x propagated to different versions like v3.10.4?

agronholm:
> Quite curious & hence one question, I see v3.10.4 is quite different branch `v3.x`. However, the only active branch is `3.x`
> How are the changes from v3.x propagated to different versions like v3.10.4?

I just create tags on specific commits on the 3.x branch, there's nothing special to it. Or did I misunderstand your question?

i-am-darshil:
Okay, so the shutdown process at `apscheduler` seems fine. Its a weird race condition that is happening at `concurrent/futures/thread.py` level.

Let me try to explain what I understand:
1. A job gets submitted [here](https://github.com/agronholm/apscheduler/blob/3.x/src/apscheduler/executors/pool.py#L27) via a pool
2. Before the thread gets `self._shutdown_lock` & `_global_shutdown_lock` [here](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L200), lets say the program exits
3. `_python_exit` acquires `_global_shutdown_lock` [here](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L25) and sets `_shutdown` to true
4. Next, [this](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L133) sets executor's `_shutdown` to True.
5. Now the thread (from point 2) gets the required locks and proceeds to encounter this condition [here](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L204) which leads to `RuntimeError('cannot schedule new futures after shutdown')`

I am not sure if there is anything we can do at `apscheduler` level apart from sweeping the error message like done in this PR.

@agronholm, your thoughts?

i-am-darshil:
Hey @agronholm, what do you think?

agronholm:
During scheduler shutdown, `ThreadPoolExecutor.shutdown(wait=True)` should have been called. After that, there shouldn't be any lingering threads around. If this is not happening, we need to check why that is.

i-am-darshil:
@agronholm 
`ThreadPoolExecutor.shutdown(wait=True)` gets called. 
I ran the code locally, and shutdown is working as expected, and no new jobs are added post shutdown. As mentioned above, its a peculiar race condition, where the job is already submitted to the `concurrent.futures.ThreadPoolExecutor` but hasn't got the cpu cycle to move forward yet. In the meantime, because of `_global_shutdown_lock` being set due to program exit, `concurrent.futures.ThreadPoolExecutor` is marked as shutdown. Then, when the job that was already to-be scheduled gets the cpu-cycle, it sees `concurrent.futures.ThreadPoolExecutor` as shutdown and raises this RuntimeError.

> 1. A job gets submitted [here](https://github.com/agronholm/apscheduler/blob/3.x/src/apscheduler/executors/pool.py#L27) via a pool
> 2. Before the thread gets `self._shutdown_lock` & `_global_shutdown_lock` [here](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L200), lets say the program exits
> 3. `_python_exit` acquires `_global_shutdown_lock` [here](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L25) and sets `_shutdown` to true
> 4. Next, [this](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L133) sets executor's `_shutdown` to True.
> 5. Now the thread (from point 2) gets the required locks and proceeds to encounter this condition [here](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/thread.py#L204) which leads to `RuntimeError('cannot schedule new futures after shutdown')`

PR #1027, opened 2025-02-20
Enhance with conf wait seconds parameter

JackLuguibin:
Enhance with conf wait seconds parameter


agronholm:
Am I supposed to understand what you meant with just that? And it looks like you targeted the wrong branch too.

Issue #967, opened 2024-09-22
App Crashed


lucasjinreal:

agronholm:
This was already fixed in e91c8ded. Would you try installing from `master` while waiting for the upcoming release?

Issue #834, opened 2024-01-01
asyncio.exceptions.CancelledError: Cancelled by cancel scope 7f27c9d9bbe0


VadyChel:
### Things to check first

- [X] I have checked that my issue does not already have a solution in the [FAQ](https://apscheduler.readthedocs.io/en/master/faq.html)

- [X] I have searched the existing issues and didn't find my bug already reported there

- [X] I have checked that my bug is still present in the latest release

### Version

4.0.0a4

### What happened?

In random moment scheduler is crashing and throw this error:
```
api      | 2024-01-01 14:22:34,081 | INFO     | apscheduler._schedulers.async_:917 - Job 01e098c6-6d31-4ad4-9cf5-f331670c11fe was cancelled
api      | 2024-01-01 14:22:34,096 | ERROR    | apscheduler._schedulers.async_:693 - Scheduler crashed
api      |   + Exception Group Traceback (most recent call last):
api      |   |   File "/usr/local/lib/python3.10/site-packages/apscheduler/_schedulers/async_.py", line 667, in run_until_stopped
api      |   |     async with create_task_group() as task_group:
api      |   |   File "/usr/local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 678, in __aexit__
api      |   |     raise BaseExceptionGroup(
api      |   | exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
api      |   +-+---------------- 1 ----------------
api      |     | Traceback (most recent call last):
api      |     |   File "/usr/local/lib/python3.10/site-packages/apscheduler/_schedulers/async_.py", line 889, in _process_jobs
api      |     |     await wakeup_event.wait()
api      |     |   File "/usr/local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 1641, in wait
api      |     |     await self._event.wait()
api      |     |   File "/usr/local/lib/python3.10/asyncio/locks.py", line 214, in wait
api      |     |     await fut
api      |     | asyncio.exceptions.CancelledError: Cancelled by cancel scope 7f27c9de4f40
api      |     |
api      |     | During handling of the above exception, another exception occurred:
api      |     |
api      |     | Exception Group Traceback (most recent call last):
api      |     |   File "/usr/local/lib/python3.10/site-packages/apscheduler/_schedulers/async_.py", line 863, in _process_jobs
api      |     |     async with AsyncExitStack() as exit_stack:
[...]

PR #1030, opened 2025-03-05
v4: Fix broken job lease extension

### LINKED ISSUE #803: APScheduler 4.0.0 Bug
### Things to check first

### Version

4.0.0a2.post72

### What happened?

I know it is an alpha etc. But I wanted to report it, because I didn't observe it with lower number of jobs:
I use the apscheduler with a postgresql jobstore and have around 5000 schedules, but the same task, just different arguments. They are scheduled with a DateTrigger and schedule themself at the end of the task for the next execution.

After some runtime some of the schedules are simply gone, the task itself had no errors and ensures in theory always a rescheduling. Meaning the number of schedules drops constantly. I observed this behavior with the thread executor but also the normal async executor. With the thread executor often the complete scheduler would crash occasionally, With the async one at least the scheduler stays alive. But both times the jobs table has an enourmous number of rows. Right now I have around 100k rows in there (I can provide a dump if needed). Also the logs shows the following error multiple times (same error just another scope valueand UUID):
```
  |   File "/root/ccn_tracker/scheduler_main.py", line 90, in main
  |     await scheduler.run_until_stopped()
  |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 514, in run_until_stopped
  |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 664, in __aexit__
  | exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/anyio/streams/memory.py", line 217, in send
    |     self.send_nowait(item)
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/anyio/streams/memory.py", line 200, in send_nowait
    |     raise WouldBlock
    | anyio.WouldBlock
    | 
    | During handling of the above exception, another exception occurred:
    | 
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 700, in _process_jobs
    |     jobs = await self.data_store.acquire_jobs(self.identity, limit)
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/datastores/sqlalchemy.py", line 759, in acquire_jobs
    |     await self._event_broker.publish(
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/eventbrokers/asyncpg.py", line 179, in publish
    |     await self._send.send(notification)
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/anyio/streams/memory.py", line 223, in send
    |     await send_event.wait()
    |   File "/usr/local/lib/python3.10/asyncio/locks.py", line 214, in wait
    |     await fut
    |   File "/usr/local/lib/python3.10/asyncio/futures.py", line 285, in __await__
    |     yield self  # This tells Task to wait for completion.
    |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 304, in __wakeup
    |     future.result()
    |   File "/usr/local/lib/python3.10/asyncio/futures.py", line 196, in result
    |     raise exc
    | asyncio.exceptions.CancelledError: Cancelled by cancel scope 7f6a84b10610
    | 
    | During handling of the above exception, another exception occurred:
    | 
    | Exception Group Traceback (most recent call last):
    |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
    |     result = coro.send(None)
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 683, in _process_jobs
    |     async with AsyncExitStack() as exit_stack:
    |   File "/usr/local/lib/python3.10/contextlib.py", line 714, in __aexit__
    |     raise exc_details[1]
    |   File "/usr/local/lib/python3.10/contextlib.py", line 697, in __aexit__
    |     cb_suppress = await cb(*exc_details)
    |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 664, in __aexit__
    | exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (82 sub-exceptions)
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 2 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 3 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 4 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 5 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 6 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 7 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 8 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 9 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 10 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 11 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 12 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 13 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 14 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- 15 ----------------
      |   File "/usr/local/lib/python3.10/asyncio/tasks.py", line 232, in __step
      |     result = coro.send(None)
      |   File "/root/.venv/ccn_tracker/lib/python3.10/site-packages/apscheduler/schedulers/async_.py", line 785, in _run_job
      |     self._running_jobs.remove(job.id)
      | KeyError: UUID('7baf47d5-46f3-4c56-b4ca-9fcdc4189250')
      +---------------- ... ----------------
      | and 67 more exceptions
      +------------------------------------
```

```
from __future__ import annotations

import asyncio
import asyncpg
import sys
import logging
import traceback
from datetime import datetime, timedelta, timezone
from logging import handlers
from typing import Optional


if sys.version_info >= (3, 11):
	from typing import Self
else:
	from typing_extensions import Self

import apscheduler
from apscheduler._schedulers.async_ import AsyncScheduler
from apscheduler.datastores.sqlalchemy import SQLAlchemyDataStore
from apscheduler.serializers.json import JSONSerializer
from sqlalchemy.ext.asyncio import create_async_engine


# logging
logger = logging.getLogger('bug_demo')
file_handler = handlers.TimedRotatingFileHandler(filename=f"bug_demo.log",
                                                 when='MIDNIGHT', encoding='utf-8',
                                                 backupCount=0, utc=False)
file_handler.setFormatter(logging.Formatter('%(asctime)s:%(levelname)s-%(name)s:\n'
                                            '%(module)s - %(funcName)s - %(lineno)d:\n'
                                            '%(message)s\n\n'))
file_handler.namer = lambda name: name.replace(".log.", "_") + ".log"
logger.addHandler(file_handler)
logger.setLevel(logging.ERROR)

# dedicated logger for apscheduler
logger2 = logging.getLogger("apscheduler")
file_handler2 = handlers.TimedRotatingFileHandler(filename=f"bug_demo_appscheduler.log",
                                                 when='MIDNIGHT', encoding='utf-8',
                                                 backupCount=0, utc=False)
file_handler2.setFormatter(logging.Formatter('%(asctime)s:%(levelname)s-%(name)s:\n'
                                            '%(module)s - %(funcName)s - %(lineno)d:\n'
                                            '%(message)s\n\n'))
file_handler2.namer = lambda name: name.replace(".log.", "_") + ".log"
logger2.addHandler(file_handler2)
logger2.setLevel(logging.INFO)


cfg = {'user'    : 'ENTER USER HERE',
       'password': 'ENTER PASSWORD HERE',
       'host'    : 'ENTER HOST HERE',
       'database': 'ENTER DATABASE HERE',
       'port': 'ENTER PORT HERE'
       }


class ExtendedAsyncScheduler(AsyncScheduler):
	async def __aenter__(self: Self) -> Self:
		await super().__aenter__()
		self.db = await asyncpg.create_pool(**cfg)
		self.logger = logger
		return self


async def regular_check(check_id: str, scheduler: Optional[ExtendedAsyncScheduler] = None):
	if not scheduler:
		try:
			scheduler: ExtendedAsyncScheduler = apscheduler.current_async_scheduler.get()
		except LookupError as e:
			raise e
	try:
		run_date = datetime.now(tz=timezone.utc) + timedelta(minutes=5)
		await scheduler.add_schedule(regular_check,
		                             DateTrigger(run_time=run_date),
		                             id=f"regular_check-{check_id}",
		                             misfire_grace_time=300,
		                             conflict_policy=apscheduler.ConflictPolicy.replace,
		                             args=(check_id,),
		                             max_jitter=30)
		logger.info(f"Added regular check {check_id} at {run_date}")
	except Exception as e:
		logger.error(f"Error in regular check {check_id}\n{traceback.format_exc()}")


async def check_self(scheduler: Optional[ExtendedAsyncScheduler] = None):
	if not scheduler:
		try:
			scheduler: ExtendedAsyncScheduler = apscheduler.current_async_scheduler.get()
		except LookupError as e:
			raise e
	db: asyncpg.Pool = scheduler.db
	try:
		async with db.acquire() as conn:
			jobs = await conn.fetchrow("SELECT count(*) as total, count(*) filter ( where jobs.task_id is not null and "
			                         "task_id ilike '__main__:regular_check' ) as reg "
			                         "from apscheduler.jobs")
			if not jobs:
				jobs = (0, 0)
			schedules = await conn.fetchrow("SELECT count(*) as total, count(*) filter ( where "
			                         "id ilike 'regular_check%' ) as req from apscheduler.schedules")
			if not schedules:
				schedules = (0, 0)
		now = datetime.utcnow()
		scheduler.logger.error(f"{now}\njobs: {jobs[0]}, schedules: {schedules[0]}, "
		                       f"regular_check schedules: {schedules[1]}, regular_check jobs: {jobs[1]}\n"
		                       f"regular_check jobs+schedules: {schedules[1] + jobs[1]}")
	except Exception as e:
		logger.error(f"Error in check_self\n{traceback.format_exc()}")

async def create_schedules(scheduler: Optional[ExtendedAsyncScheduler] = None):
	if not scheduler:
		try:
			scheduler: ExtendedAsyncScheduler = apscheduler.current_async_scheduler.get()
		except LookupError as e:
			raise e
	db: asyncpg.Pool = scheduler.db
	try:
		async with db.acquire() as conn:
			for i in range(2000):
				try:
					schedule = await conn.fetchrow("SELECT * from apscheduler.schedules where id ilike $1",
					                              f"regular_check-{i}")
				except Exception as e:
					traceback.print_exc()
					raise e
				if not schedule:
					await scheduler.add_job(regular_check, args=(f'{i}',))
		logger.error("Finished creating schedules")
	except Exception as e:
		logger.error(f"Error in create_schedules\n{traceback.format_exc()}")

async def main():
	engine = create_async_engine(
			f"postgresql+asyncpg://{cfg['user']}:{cfg['password']}@{cfg['host']}/{cfg['database']}")
	serializer = JSONSerializer()
	data_store = SQLAlchemyDataStore(engine, serializer=serializer, schema='apscheduler')
	async with ExtendedAsyncScheduler(data_store=data_store,
	                                  role=apscheduler.SchedulerRole.both) as scheduler:
		await check_self(scheduler)
		await scheduler.add_schedule(check_self,
		                             IntervalTrigger(minutes=1),
		                             id="check_self",
		                             misfire_grace_time=120)
		# await scheduler.add_job(create_schedules) # creates bug directly before scheduler even starts
		await create_schedules(scheduler) # works fine for the first few minutes and then the bug occurs again
		await check_self(scheduler)
		print('starting scheduler')


if __name__ == '__main__':
	asyncio.run(main())

```

JacobHayes:
## Changes

The `extend_job_leases` function was checking for `RunState.started` in a `while` loop, however the coroutine is first executed by the (already running) job executors when the scheduler is still `RunState.starting`. This caused the loop condition to immediately evaluate to `False` and the function to return. This commit fixes that by checking for both `RunState.starting` and `RunState.started` in `extend_job_leases`, similar to the `cleanup` task.

With the `MemoryDataStore`, if a job ran longer than the lease *and* past the next scheduled execution, then the exact same job would be executed again (along with a job for the next schedule). The first execution cleans up when it finally finishes, but then the second execution fails in cleanup.

The issue primarily affects the `MemoryDataStore` because its `acquire_jobs` method seems to acquire even expired jobs (which the scheduler re-executes) more aggressively than the other data stores, which have different logic for discarding jobs / handling task job slots.

This commit partially fixes #803, #834, #952, #967, and maybe others. I suspect there are more bugs lurking that I haven't pinned down yet, likely related to the acquiring, releasing, and perhaps cleanup logic.

---

Do you want me to add the repro from https://github.com/agronholm/apscheduler/issues/952#issuecomment-2695252917 as a test, ensuring it doesn't fail after say 10 seconds or something? It's admittedly a bit contrived (and hard to prove a negative), but I think represents the same issue I've been seeing with longer timespans.

## Checklist

features)

these instructions.

### Updating the changelog


agronholm:
I'll have to take my time evaluating this, first refreshing my memory about the v4 code base and how the lease extensions work. I'll try to do it this week.

JacobHayes:
Thanks! I'm not sure how/if the lease extension ever worked, even in the initial commit (https://github.com/agronholm/apscheduler/commit/e91c8ded7e848ad2b20c652a50b7a729677c7b3d) the `extend_job_leases` task was scheduled (and _likely_ executed by the running executors) before the `await self.get_next_event(SchedulerStarted)`.

One easy quick way to check this is just by adding a `print("Ending extend_job_leases")` after the `while` loop in `extend_job_leases` with and without this PR. Without the PR, it just exits immediately (and nothing else ever schedules it again).

agronholm:
> FYI I'm testing this in prod on a small/non-critical app which was triggering the bug ~10 times a day. I'll report in a couple of days with the output.

You're quite brave to run an alpha version in production!

HK-Mattew:
> You're quite brave to run an alpha version in production!

There are some crazy people following you hahah.

Well, I'm also using it in production and honestly, for me, there are no more bugs. I've had APScheduler V4 processes running for over 3 months and not a single error has occurred (and it's working very well).

I believe that for this lib to come out as a stable version, there are only a few things missing.

agronholm:
OK, so I've taken an initial look and refreshed my memory about this a bit. While I think you've correctly recognized that the loop will exit prematurely, the underlying issues with this functionality run a bit deeper. What needs to happen when the scheduler is starting is that any jobs left in the running state by this scheduler should be reaped, marked as `abandoned`. This is done in a separate maintenance loop (`AsyncScheduler._cleanup_loop`). What's missing, I think, is that right when the scheduler starts, it should clean up any such leftover jobs for that particular scheduler, even if the lease duration hasn't expired yet. This step should be done before the first run of `extend_acquired_job_leases()`. I will work on a proper fix for this.

JacobHayes:
To confirm I'm following, you're saying we need to avoid extending the leases of jobs leftover from other (presumably dead) schedulers, correct? Or, are you referring to the same scheduler object that is stopped and then later restarted? (I'm admittedly a bit fuzzy on the multiple scheduler scenario since I've mostly dealt with the memory data store with 1 start of the scheduler per process.)

Thinking out loud a bit, since `extend_acquired_job_leases' only looks at the jobs in the local object's `self._running_jobs`:
- it wouldn't affect jobs from previous scheduler objects (leaving them for cleanup, though not until after the normal lease ends)
- could we just clear `self._running_jobs` when the scheduler starts up? On the first start, it'll obviously be empty and on the second start, we'd clear it (leaving cleanup to handle them, as above) so we only extend leases for newly started jobs.

---

Separately, another thought I had was that perhaps instead of a general "extend leases" task, we could create one as a sibling of each job task in the same TaskGroup. It's probably more tasks overall assuming you have concurrent jobs, but no tasks between jobs since it'd have the same lifetime due to normal cancelation semantics and would allow tuning the lease duration per task (if that's desirable).

agronholm:
> To confirm I'm following, you're saying we need to avoid extending the leases of jobs leftover from other (presumably dead) schedulers, correct? Or, are you referring to the same scheduler object that is stopped and then later restarted? (I'm admittedly a bit fuzzy on the multiple scheduler scenario since I've mostly dealt with the memory data store with 1 start of the scheduler per process.)
> 
> Thinking out loud a bit, since `extend_acquired_job_leases` only looks at the jobs in the local object's `self._running_jobs`:
> 
>     * it wouldn't affect jobs from previous scheduler objects (leaving them for cleanup, though not until after the normal lease ends)
> 
>     * could we just clear `self._running_jobs` when the scheduler starts up? On the first start, it'll obviously be empty and on the second start, we'd clear it (leaving cleanup to handle them, as above) so we only extend leases for newly started jobs.

I think I got confused myself, I forgot how `extend_acquired_job_leases()` worked. If the app crashed, then `self._running_jobs` would indeed be empty. But we still need some kind of operation to deal with any schedules and jobs that were left in an acquired state by the scheduler before it crashed. Perhaps this should be handled in the `start()` method? I'd need to change the signature to pass in the scheduler ID for that to work.

> Separately, another thought I had was that perhaps instead of a general "extend leases" coroutine, we could create one as a sibling of each job coroutine in the same TaskGroup. It's probably more coroutines overall assuming you have concurrent jobs, but no extend steps between jobs since it'd have the same lifetime due to normal cancelation semantics and would allow tuning the lease duration per task (if that's desirable).

I read this paragraph twice and was still left confused. What is it that you're actually proposing?

agronholm:
I've implemented the purging of previously acquired schedules and jobs by the same scheduler in the `start()` method of both SQLAlchemy and MongoDB data stores (this is irrelevant to memory job stores). Now I just need to write the appropriate tests.

agronholm:
I opted to add this as two new abstract methods in `DataStore`. I have data store level tests for this but I want to add scheduler level tests too to make sure it doesn't break.

JacobHayes:
> But we still need some kind of operation to deal with any schedules and jobs that were left in an acquired state by the scheduler before it crashed.
> [...]
> I've implemented the purging of previously acquired schedules and jobs by the same scheduler in the start() method of both SQLAlchemy and MongoDB data stores

I thought `AsyncScheduler.cleanup` / `DataStore.cleanup` did something similar now, though they would wait for prior `lease_duration`s on the acquired resources to end. Are these methods different in that they ignore that to clean up right away?

That does seem like an improvement if the scheduler crashed and gets restarted immediately (so it doesn't hang until the previous leases expire), but does that mean if a new scheduler with the same identity starts before an old one winds down (eg: a rolling deployment) that it would kill the other's jobs? I'm guessing there are other issues when two schedulers with shared identity overlap (and the default is ~uniquely generated), so perhaps that's not quite worth considering.

agronholm:
> > But we still need some kind of operation to deal with any schedules and jobs that were left in an acquired state by the scheduler before it crashed.
> > [...]
> > I've implemented the purging of previously acquired schedules and jobs by the same scheduler in the start() method of both SQLAlchemy and MongoDB data stores
> 
> I thought `AsyncScheduler.cleanup` / `DataStore.cleanup` did something similar now, though they would wait for prior `lease_duration`s on the acquired resources to end. Are these methods different in that they ignore that to clean up right away? Also, this only applies if the `scheduler.identity` is set explicitly (since it's otherwise ~uniquely generated so wouldn't match after a crash), right?
> 
> That does seem like an improvement if the scheduler crashed and gets restarted immediately (so it doesn't hang until the previous leases expire), but does that mean if a new scheduler with the same identity starts before an old one winds down (eg: a rolling deployment) that it would kill the other's jobs?
> 
> I'm guessing there are other issues when two schedulers with shared identity overlap, so perhaps that's not quite worth considering.

The way scheduler identities are automatically generated virtually guarantees uniqueness. But to answer your question, if a new scheduler was starting with the same identity as an older one that is still running, bad things would happen (it would mark already running jobs as abandoned).

So yes, the idea is that a scheduler, on startup, cleans up purges any records of running jobs and acquired schedules marked to its own scheduler ID because no other scheduler can do it until the leases expire, and that way the delay is minimized.

JacobHayes:
Update: I haven't had any more scheduler crashes since using this PR for the last month (vs ~1/day before this PR).

Any opposition to merging this PR in?

---

On the other cleanup stuff you mentioned:

> The way scheduler identities are automatically generated virtually guarantees uniqueness.
> [...]
> So yes, the idea is that a scheduler, on startup, cleans up purges any records of running jobs and acquired schedules marked to its own scheduler ID because no other scheduler can do it until the leases expire, and that way the delay is minimized.

I think my confusion is this: if the scheduler ID _is_ unique, then when would there be running jobs or acquired schedules marked to the scheduler's ID on startup?

agronholm:
> Update: I haven't had any more scheduler crashes since using this PR for the last month (vs ~1/day before this PR).
> 
> Any opposition to merging this PR in?

I had my own local changes (which do more than what you've done here) almost ready to go, just waiting for some additional tests. Sorry for dropping the ball!

> On the other cleanup stuff you mentioned:
> 
> > The way scheduler identities are automatically generated virtually guarantees uniqueness.
> > [...]
> > So yes, the idea is that a scheduler, on startup, cleans up purges any records of running jobs and acquired schedules marked to its own scheduler ID because no other scheduler can do it until the leases expire, and that way the delay is minimized.
> 
> I think my confusion is this: if the scheduler ID _is_ unique, then when would there be running jobs or acquired schedules marked to the scheduler's ID on startup?

If you launch the same (or another scheduler) with the same ID. Remember that the scheduler ID is only autogenerated _by default_ and can be set explicitly. This is ideal so that the scheduler can then clean up any jobs it was running while it crashed. No other scheduler could do that, as they wouldn't know that the scheduler running them has crashed before the jobs expire.

JacobHayes:
> I had my own local changes (which do more than what you've done here) almost ready to go, just waiting for some additional tests. Sorry for dropping the ball!

No worries, just checking there's nothing else on my end. :)

> If you launch the same (or another scheduler) with the same ID.

If you did set the same ID, how do we know that the _previous_ scheduler is indeed dead (and thus it is safe to clean up their jobs)?

If the previous one isn't dead, this cleanup would mask a likely configuration issue (not-unique-enough scheduler ID) with subtle issues:
- failure to cleanup in the first process after the second process has cleared its jobs
- duplicate runs of a job when the second process clears and then start jobs the first process is still running

A common scenario I'm imagining is a rolling deployment where a new process is spun up _before_ the previous one is shut down. In that case, if the user is setting a static scheduler ID (which could otherwise be fine and aid crash recovery as you mention), the second process would _incorrectly_ (imo) clean up jobs of the still-running previous process that hasn't shut down yet.

---

Unless I’m missing something else that prevents the above (very possible 😁), perhaps we could introduce a “scheduler heartbeat” mechanism that:
- tracks the last time a given scheduler ID was seen
- only allows clean up after a small "heartbeat timeout" has elapsed without any updates
    - allowing ~fast recovery-after-a-crash, independent of `lease_duration`
- errors loudly if a duplicate scheduler sees an active heartbeat for the same ID
    - clearly highlighting a likely configuration issue

That may still have some false positives if for whatever reason a scheduler can't check in (eg: a blocked async task, a slow task that doesn't release the GIL in the thread executor, etc), but I think would handle the common error cases more cleanly for users.

In addition to "scheduler_id" and "last_seen_at", it could also track a randomly generated "token" that is pinned to a specific Scheduler object so you can handle data races with a compare-and-swap if eg: 2+ other schedulers start all with the same ID, both see a dead heartbeat, and then both try to claim it.

I have a couple more thoughts, but will hold off to avoid being _even more_ long winded until/if you think it's worth looking into. :)

agronholm:
> If you did set the same ID, how do we know that the previous scheduler is indeed dead (and thus it is safe to clean up their jobs)?

You should only do use a static ID for your scheduler when you **know** it will never be duplicated. A practical example of this is a Kubernetes StatefulSet where each pod retains its specific identity that will never be duplicated by another pod.

> Unless I’m missing something else that prevents the above (very possible 😁), perhaps we could introduce a “scheduler heartbeat” mechanism that:

The mechanism I already have in place here is similar, but it allows for other schedulers to reap jobs abandoned by another scheduler, provided that the "heartbeat" failed (that is, the job's `acquired_until` timestamp has passed. But if a scheduler with a static ID starts up, it *knows* that any jobs in the data store marked with its own ID are definitely dead, and can be reaped immediately.

Does this make more sense now?

agronholm:
I'll push my own changeset some time over the weekend. It should further clarify my intentions.

agronholm:
Today I'm working on the code again, and figured out an alternate way for schedules at least. But strangely the scheduler-level test I wrote is already passing and it shouldn't be.

agronholm:
Ah, that turned out to be a simple mistake in how I subscribed to events! Now it makes sense.

agronholm:
I've pushed the schedule related changes. The reaping of abandoned jobs will follow.

agronholm:
The reaping of abandoned jobs is in now. Next I'll check if this PR is still needed.

agronholm:
Thanks.

JacobHayes:
Thanks for answering all my questions here and the thorough checks!

PR #1031, opened 2025-03-06
Skip dispatching extend_acquired_job_leases with no jobs

JacobHayes:
## Changes

After fixing https://github.com/agronholm/apscheduler/pull/1030, I noticed that we'd schedule the lease extension tasks even though there were no jobs to extend. This is a small "optimization" to skip creating those no-op tasks.

This isn't really visible without https://github.com/agronholm/apscheduler/pull/1030 since the `while` loop never hit without that fix, but GH doesn't support stacked PRs from forks so figured I'd push separately :shrug:.

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -64,8 +64,7 @@ APScheduler, see the :doc:`migration section <migration>`.
   schema if it's missing (PR by @zhu0629)
 - Fixed an issue with ``CronTrigger`` infinitely looping to get next date when DST ends
   (`#980 <https://github.com/agronholm/apscheduler/issues/980>`_; PR by @hlobit)
-- Skip dispatching extend_acquired_job_leases with no jobs (`#1031
-  <https://github.com/agronholm/apscheduler/issues/1031>`_; PR by @JacobHayes)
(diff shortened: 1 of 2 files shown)


docs/versionhistory.rst
```diff
@@ -64,6 +64,8 @@ APScheduler, see the :doc:`migration section <migration>`.
   schema if it's missing (PR by @zhu0629)
 - Fixed an issue with ``CronTrigger`` infinitely looping to get next date when DST ends
   (`#980 <https://github.com/agronholm/apscheduler/issues/980>`_; PR by @hlobit)
+- Skip dispatching extend_acquired_job_leases with no jobs (`#1031
+  <https://github.com/agronholm/apscheduler/issues/1031>`_; PR by @JacobHayes)
```
agronholm, line 68: The link is invalid, and PR links are not needed.
```suggestion
- Skip dispatching extend_acquired_job_leases with no jobs (PR by @JacobHayes)
```

agronholm (approved):
Looks reasonable!

PR #1039, opened 2025-04-11
Store `last_fire_time` on schedule release

hlobit:
I noticed that the value of `last_fire_time` (present in the `ScheduleResult`) is not stored in any of the available datastore implementations.

Here is an attempt to store the value.

### DIFF
--- a/src/apscheduler/datastores/sqlalchemy.py
+++ b/src/apscheduler/datastores/sqlalchemy.py
@@ -118,7 +118,8 @@ def marshal_timestamp(timestamp: datetime | None, key: str) -> Mapping[str, Any]
 
     return {
         key: int(timestamp.timestamp() * 1000_000),
-        key + "_utcoffset": cast(timedelta, timestamp.utcoffset()).total_seconds() // 60,
+        key + "_utcoffset": cast(timedelta, timestamp.utcoffset()).total_seconds()
+        // 60,
     }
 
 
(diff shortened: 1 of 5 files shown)


agronholm:
Looks promising! A changelog entry is needed. Check the existing entries and add your own in the same style in `docs/versionhistory.rst`.

hlobit:
@agronholm changelog entry added!

Though I cannot say if it could be considered as a breaking change.
Maybe for aiosqlite/SQLAlchemy? Isn't a new column needed for this datastore?

src/apscheduler/datastores/sqlalchemy.py
```diff
@@ -257,37 +268,46 @@ def _temporary_failure_exceptions(self) -> tuple[type[Exception], ...]:
...
+                else:
+                    data[f"{field}_utcoffset"] = None
 
         return data
 
     def get_table_definitions(self) -> MetaData:
         if self._supports_tzaware_timestamps:
             timestamp_type: TypeEngine[datetime] = DateTime(timezone=True)
+            last_fire_time_tzoffset_columns: tuple[Column, ...] = (
+                Column("last_fire_time", timestamp_type, index=True),
```
agronholm, line 300: Same as above.

agronholm (changes requested):
Looking good otherwise, but maybe drop that unnecessary new index?

agronholm:
WDYM, new column?

> @agronholm changelog entry added!
> 
> Though I cannot say if it could be considered as a breaking change. Maybe for aiosqlite/SQLAlchemy? Isn't a new column needed for this datastore?

Hm, 

> @agronholm changelog entry added!
> 
> Though I cannot say if it could be considered as a breaking change. Maybe for aiosqlite/SQLAlchemy? Isn't a new column needed for this datastore?

Breaking changes are fine, given that this is still in alpha.

agronholm (approved):
LGTM! Thanks.

PR #1042, opened 2025-04-19
Refactor example to use FastAPI lifespan instead of custom middleware

dzherb:
## Changes
This PR updates the example to use the FastAPI `lifespan` context manager instead of relying on the `scope` middleware pattern for startup and shutdown events. This approach is more idiomatic and easier to understand.

Specifically:

- Replaced the custom `SchedulerMiddleware` with a `lifespan` context.

- Moved scheduler setup into the lifespan handler.

- Cleaned up unnecessary middleware and request handling code.

Let me know if you'd like me to keep the old version as a separate example or add any additional explanation in comments or docs :)

### DIFF
--- a/examples/web/asgi_fastapi.py
+++ b/examples/web/asgi_fastapi.py
@@ -11,14 +11,13 @@ basic web app at http://localhost:8000.
 
 from __future__ import annotations
 
+from collections.abc import AsyncGenerator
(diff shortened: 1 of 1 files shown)


agronholm (approved):

agronholm:
Thanks.

PR #1054, opened 2025-06-06
[fix]]Add a argument `create_table` on class `SQLAlchemyJobStore`

Dontes-cn:
## Changes
Add a argument `create_table` on class `SQLAlchemyJobStore`

Fixes #. 

I use a db which not supposed "if not exists",when i start the apscheduler job,it has an exception

sqlalchemy.exc.DatabaseError:(dmPython.DatabaseError) [CODE:-2124]Error in line: 9
Object [apscheduler jobs] already exists
[SQL:
CREATE TABLE apscheduler jobs(
id NVARCHAR2(191)NOT NULL,next run time FLOAT.
job state BLOB NOT NULL,
PRIMARY KEY (id)
)

## Checklist

- [√ ] You've updated the documentation (in `docs/`, in case of behavior changes or new
features)
- [√ ] You've added a new changelog entry (in `docs/versionhistory.rst`).

these instructions.

### Updating the changelog

### DIFF
--- a/docs/modules/jobstores/sqlalchemy.rst
+++ b/docs/modules/jobstores/sqlalchemy.rst
@@ -6,7 +6,7 @@
 API
 ---
 
-.. autoclass:: SQLAlchemyJobStore(url=None, engine=None, tablename='apscheduler_jobs', metadata=None, pickle_protocol=pickle.HIGHEST_PROTOCOL)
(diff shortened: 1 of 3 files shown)


agronholm:
SQLAlchemy only creates tables if they don't exist. What database dialect are you using? This seems like an upstream bug.

agronholm:
See [here](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.MetaData.create_all) for reference.

PR #1060, opened 2025-07-01
Fix dst skip hour

jonasitzmann:
## Changes

Fixes #. 

## Checklist

features)

these instructions.

### Updating the changelog

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,6 +6,8 @@ APScheduler, see the :doc:`migration section <migration>`.
 
 **4.0.0a6**
 
+- Fixed an issue where `CronTrigger.next` returned a non-existing date on a DST change
+  (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_)
 - **BREAKING** Refactored ``AsyncpgEventBroker`` to directly accept a connection string,
   thus eliminating the need for the ``AsyncpgEventBroker.from_dsn()`` class method
 - **BREAKING** Added the ``extend_acquired_schedule_leases()`` data store method to
--- a/src/apscheduler/triggers/cron/__init__.py
+++ b/src/apscheduler/triggers/cron/__init__.py
@@ -6,6 +6,7 @@ from typing import Any, ClassVar
 
 import attrs
 from attr.validators import instance_of, optional
+from dateutil.tz import tz
 from tzlocal import get_localzone
 
 from ..._converters import as_aware_datetime, as_timezone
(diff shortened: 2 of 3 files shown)


docs/versionhistory.rst
```diff
@@ -6,6 +6,8 @@ APScheduler, see the :doc:`migration section <migration>`.
 
 **4.0.0a6**
 
+- Fixed an issue where `CronTrigger.next` returned a non-existing date on a DST change
+  (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_)
```
agronholm, line 10: Add this under `**UNRELEASED**` as per the PR template instructions. Also code literals need two backticks on RST.
```suggestion
**UNRELEASED**

- Fixed an issue where ``CronTrigger.next()`` returned a non-existing date on a DST change
  (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_)

**4.0.0a6**
```

agronholm:
I would very much like to avoid adding a dependency just for a bug fix.

jonasitzmann:
`CronTrigger._increment_field_value` deals with this implicitly by setting
```python
dateval = datetime.fromtimestamp(dateval.timestamp() + difference.total_seconds(), self.timezone)
```
The resulting dateime object always exists.

However, it is not sufficient to convert the result of `_increment_field_value` in the same way because different fields need to checked next, depending on the existence of `next_time`.

Instead, the timestamp can be used to define a function `time_exists` as:
```python
def time_exists(dt: datetime) -> bool:
    existing_dt = datetime.fromtimestamp(dt.timestamp(), dt.tzinfo)
    return dt == existing_dt
```

agronholm:
> `CronTrigger._increment_field_value` deals with this implicitly by setting
> 
> ```python
> dateval = datetime.fromtimestamp(dateval.timestamp() + difference.total_seconds(), self.timezone)
> ```
> 
> The resulting dateime object always exists.
> 
> However, it is not sufficient to convert the result of `_increment_field_value` in the same way because different fields need to checked next, depending on the existence of `next_time`.
> 
> Instead, the timestamp can be used to define a function `time_exists` as:
> 
> ```python
> def time_exists(dt: datetime) -> bool:
>     existing_dt = datetime.fromtimestamp(dt.timestamp(), dt.tzinfo)
>     return dt == existing_dt
> ```

Oh, I think I remember now. When I used `pytz` in the 3.x series, it would raise an error if I tried to use `tz.normalize()` on a nonexistent date. But `pytz` is a thing of the past, so some other approach is needed regardless.

agronholm (approved):

PR #1062, opened 2025-07-02
Fix start end time timezone

### LINKED ISSUE #1061: CronTrigger does not convert `start_time` to `self.timezone`
### Things to check first

### Version

4.0.0a6

### What happened?

When initializing a `CronTrigger` with `timezone != start_time.tzinfo` it does not convert `start_time` to the given timezone.
This causes incorrect fire times in `CronTrigger.next()`

```python
from datetime import timezone, timedelta, datetime

def test_timezone_change() -> None:
    est = timezone(offset=timedelta(hours=-5))
    cst = timezone(offset=timedelta(hours=-6))
    start_time = datetime(2009, 9, 26, 10, 16, tzinfo=cst)
    trigger = CronTrigger(hour=11, minute="*/5", timezone=est, start_time=start_time)
    correct_next_time = datetime(2009, 9, 26, 11, 20, tzinfo=est)
    next_time = trigger.next()
    assert str(next_time) == str(correct_next_time)
```

jonasitzmann:
## Changes

Fixes #1061

- convert `start_time` and (if not None) `end_time` to `self.timezone`
- use `self.timezone` as default, when `start_time` or `end_time` are not aware

## Checklist

- (N/A) You've updated the documentation (in `docs/`, in case of behavior changes or new
features)

these instructions.

### Updating the changelog

- Fixed an issue where ``CronTrigger`` does not convert ``start_time`` to ``self.timezone``
  (`#1061 <https://github.com/agronholm/apscheduler/issues/1061>`_; PR by @jonasitzmann)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.0
+    rev: v0.12.1
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -10,6 +10,7 @@ APScheduler, see the :doc:`migration section <migration>`.
   (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_)
 
 **4.0.0a6**
+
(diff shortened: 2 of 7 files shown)


jonasitzmann:
I replaced the converter `as_aware_datetime` with a new one `as_datetime`.
This way, `start_time` and `end_time` can be made aware using `self.timezone` as default instead of `get_localzone`

jonasitzmann:
Thanks for your review! :eyes:

PR #1074, opened 2025-08-16
Add: Run `async` coroutines in `*Pool` executors

sneakers-the-rat:
## Changes

Allow `async` functions to be used in the thread and process pool executors.

Of course, for those that are running only/mostly async jobs, one could/should use the `AsyncIOExecutor`,
but we encountered a situation where we want to run pools of jobs in multiple processes,
some of which are async and some of which are not.

We had been using the AsyncIOScheduler and executor happily until now,
and I figured that we would be able to swap in/combine a ProcessPoolExecutor with the same set of jobs,
but alas, it appears as if the Pool executors don't await coroutines.

This may seem like an odd thing to add for the ThreadPoolExecutor, as each async coroutine
would be run in a separate thread and thus not be able to take advantage of asyncio's cooperative concurrency,
but I think this still makes sense from an API consistency POV - it should be possible to swap around executors
with the same set of jobs.
It also makes sense in cases like ours where the jobs use asyncio internally, e.g. one job that does a bunch of async
network calls, so it *does* make use of the event loop internally, even if it doesn't share concurrency *across* jobs like the AsyncIOExecutor does.

I tried to reuse as much as i could from within the package and match patterns that I saw,
let me know if any of this is unclear :)

I'll hold off updating the docs until i get word on whether this change would be welcome or not, sorry for not raising an issue first, i just see that several others have been made in the past surrounding this issue, and the change was small enough that it wouldn't be that big of a loss for me if y'all decide that this isn't welcome and just close it.

## Checklist

features) - will do after hearing back from maintainers <3
- [ ] You've added a new changelog entry (in `docs/versionhistory.rst`) - will do after hearing back from maintainers <3

### DIFF
--- a/src/apscheduler/executors/pool.py
+++ b/src/apscheduler/executors/pool.py
@@ -1,9 +1,11 @@
+import asyncio
 import concurrent.futures
 import multiprocessing
 from abc import abstractmethod
 from concurrent.futures.process import BrokenProcessPool
 
-from apscheduler.executors.base import BaseExecutor, run_job
+from apscheduler.executors.base import BaseExecutor, run_coroutine_job, run_job
+from apscheduler.util import iscoroutinefunction_partial
 
 
 class BasePoolExecutor(BaseExecutor):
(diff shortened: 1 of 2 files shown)


agronholm:
The 3.x branch is not getting any new features. All feature development happens in the 4.x series (`master`). If you feel that your needs are not addressed there, feel free to send your pitch with the proper justifications as a feature request.

sneakers-the-rat:
My bad. Didn't realize 3 was frozen. I'll make this change locally and wait for 4 to stabilize in that case. Thanks for this package! We love it.

agronholm:
I would still be interested to learn if v4 has the capabilities you need.

PR #1079, opened 2025-09-27
Fixes #1021. cast to ISO format to advance through fold

### LINKED ISSUE #1021: CronTrigger runs into infinite loop at daylight savings time boundary
### Things to check first

### Version

3.11.0

### What happened?

Similar bug as https://github.com/agronholm/apscheduler/issues/980 which was fixed on master, this is on v3.11. I [commented](https://github.com/agronholm/apscheduler/issues/980#issuecomment-2499977654) this in that issue, but am now creating a separate issue since that issue was closed.

```
from zoneinfo import ZoneInfo
from datetime import datetime

tz = ZoneInfo("Europe/Helsinki")
trigger = CronTrigger(minute=30, timezone=tz)
dt = datetime(2017, 10, 29, 3, 45, tzinfo=tz, fold=1)

dt = trigger.get_next_fire_time(dt, dt)
# dt = datetime.datetime(2017, 10, 29, 3, 30, fold=1, tzinfo=zoneinfo.ZoneInfo(key='Europe/Helsinki'))

dt = trigger.get_next_fire_time(dt, dt)
# dt = datetime.datetime(2017, 10, 29, 3, 30, fold=1, tzinfo=zoneinfo.ZoneInfo(key='Europe/Helsinki'))
```

berianjames:
## Changes

Fixes #1021. 

This resolves CronTrigger becoming stuck on the fall-back DST transition in the 3.x branch (already resolved in prerelease).

This solution comes from noticing that a folded datetime looks correct when rendered in its ISO format (e.g. `print(start_date)`). Casting it to the ISO format effectively 'unfolds' the datetime object into an unambiguous representation that standard time calculations can reliably advance. (This comes with the temporary change of tzinfo to be a timedelta rather than a ZoneInfo, which is added back when `next_date` is calculated.)

## Checklist

- [x] I've added tests (in `tests/`) added which would fail without my patch
- [x] I've run the tests locally with success after patching the code.
- [x] I've added a new changelog entry (in `docs/versionhistory.rst`).

No other updates to docs, because this fix doesn't change feature behaviour.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -8,7 +8,7 @@ APScheduler, see the :doc:`migration section <migration>`.
 
 - Fixed ``scheduler.shutdown()`` not raising ``SchedulerNotRunning`` (or raising the
   wrong exception) for asynchronous schedulers when the scheduler is in fact not running
-- Fixed ``CronTrigger`` sticking on a folded datetime during the fall-back DST transition.
+- Fixed ``CronTrigger`` sticking on a folded datetime during the fall-back DST transition
   (#1021 <https://github.com/agronholm/apscheduler/issues/1021>_; PR by @berianjames)`
 
 **3.11.0**


Pwuts:
That would be amazing, thank you so much :)

agronholm:
It's a rather ugly solution but the test proves that it works. If I can't think of anything better, I'll merge it soon.

agronholm:
Sure, I'll do that today.

berianjames:
Great work, thanks Alex

Issue #123, opened 2016-04-05
'tuple' object has no attribute 'public' in __init__ 


ajain352:
As i can see the commit which added this piece of code in constructor was added just 3 days back. So I am just curious if its a bug or I am doing something wrong here ?
Trace:
      2 parsed_version = __import__('pkg_resources').get_distribution('APScheduler').parsed_version
----> 3 version_info = tuple(int(x) if x.isdigit() else x for x in parsed_version.public.split('.'))
      4 version = parsed_version.base_version
      5 release = __version__ = parsed_version.public

AttributeError: 'tuple' object has no attribute 'public'

backbord:
Hi, for me this caused problems too, when I was using setuptools 18.x.y.
With a current version of setuptools (20.6.7), it works fine.

agronholm:
Damnit, I had no idea this would break on older setuptools.

agronholm:
@ajain352 what exact version of setuptools are you using? What about you @backbord ? I couldn't reproduce this even with setuptools 9.1.

backbord:
Hi, I've had the same exact problem as ajain352 with setuptools 0.9.1 and 18.3.2.

agronholm:
I just installed setuptools 18.3.2 and "import apscheduler" works fine. Please verify.

backbord:
Oops. Wrong branch. 18.3.2 worked fine. 0.9.1 didn't work. 20.6.7 works. Haven't tried versions other than those.

wowkin2:
The same problem (Windows 7 x64, Python 3.4.3, APScheduler 3.1.0):
`Traceback (most recent call last):
  File "....../main.py", line 10, in <module>
    from apscheduler.schedulers.background import BackgroundScheduler
  File "............\venv\lib\site-packages\apscheduler\__init__.py", line 3, in <module>
    version_info = tuple(int(x) if x.isdigit() else x for x in parsed_version.public.split('.'))
AttributeError: 'tuple' object has no attribute 'public'`

agronholm:
Not sure why you're commenting this in a closed issue. A fix has been made and will appear in the next release that will be out soon.

agronholm:
Meanwhile upgrade your setuptools manually and it will work fine.

agronholm:
An alternative fix has been committed which does not require setuptools to be upgraded.

kalman770:
[...]

PR #1094, opened 2025-12-17
Add fix for get_next_fire_time not advancing through fold with unfolded previous_fire_time

### LINKED ISSUE #1021: CronTrigger runs into infinite loop at daylight savings time boundary

SoulofAkuma:
## Changes

Fixes #1021. 

As described in #1021 , there has been an issue with a Job for `CronTrigger`s during DST changes in case a `Job` has a `Job.next_run_time` with `datetime.fold == 0` while `now` has `datetime.fold == 1`. This is because the Python `datetime` comparison is not `fold` aware/ignores the attribute for datetime comparisons. This includes the `min` and `max` calls in `CronTrigger.get_next_fire_time`.

I have also removed the previous workaround converting between iso formats, because it produced invalid datetimes where the fold argument was not set according to the actual timezone offset, date and time. With this fix, this workaround is no longer necessary.

Additionally, I have adjusted a previous test case, because it tested for newly introduced, from my understanding (though please correct me if I'm wrong) invalid behavior. The previous workaround introduced the following behavior through adding one microsecond only to `now` inputs which where folded:

```py
# Example for next fire time when now is folded
# advances to next hour although the next fire time would be the input
timezone = ZoneInfo("Europe/Helsinki")
trigger = CronTrigger(minute=30, timezone=timezone)
start_date = datetime(2017, 10, 29, 3, 30, tzinfo=timezone, fold=1)
correct_next_date = datetime(2017, 10, 29, 4, 30, tzinfo=timezone, fold=0)
assert str(trigger.get_next_fire_time(None, start_date)) == str(correct_next_date)

# Example for next fire time when now is not folded 
# does not advance to next hour because the next fire time is the input
timezone = ZoneInfo("Europe/Helsinki")
trigger = CronTrigger(minute=30, timezone=timezone)
start_date = datetime(2025, 12, 1, 3, 30, tzinfo=timezone)
correct_next_date = datetime(2025, 12, 1, 3, 30, tzinfo=timezone)
assert str(trigger.get_next_fire_time(None, start_date)) == str(correct_next_date)
```

## Checklist

features) - No new features added

these instructions.

I will add all of these later and will also remove one of the commented out test cases which I wanted to highlight in the PR.

### Updating the changelog

`* Fix big bad boo-boo in the async scheduler

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -6,8 +6,9 @@ APScheduler, see the :doc:`migration section <migration>`.
 
 **UNRELEASED**
 
-- Fixed an issue where ``Job``s scheduled in a repeated time interval during DST
-  transitions could cause the scheduler to get stuck in an infinite loop
+- Fixed an issue where a job using a ``CronTrigger`` scheduled in a repeated time
+  interval during DST transitions could cause the scheduler to get stuck in an infinite
+  loop
   (#1021 <https://github.com/agronholm/apscheduler/issues/1021>_; PR by @soulofakuma)
 
 **3.11.1**


tests/triggers/test_cron.py
```diff
@@ -265,31 +265,61 @@ def test_different_tz(timezone):
 
 
 @pytest.mark.parametrize(
-    "trigger_args, start_date, start_date_fold, correct_next_date",
+    "trigger_args, start_date, start_date_fold, correct_next_date, correct_next_date_fold",
     [
-        ({"hour": 8}, datetime(2013, 3, 9, 12), False, datetime(2013, 3, 10, 8)),
-        ({"hour": 8}, datetime(2013, 11, 2, 12), True, datetime(2013, 11, 3, 8)),
+        ({"hour": 8}, datetime(2013, 3, 9, 12), 0, datetime(2013, 3, 10, 8), 0),
+        # ({"hour": 2, "minute": 30}, datetime(2013, 3, 9, 2, 31), 0, datetime(2013, 3, 11, 2, 30), 0),
```
SoulofAkuma: I left this test case in here, because it is currently failing and from my understanding a different bug of `CronTrigger`.

If an time does not exist in a day, because that time is skipped, I would expect that `CronTrigger` to run on that day.
However, `CronTrigger` does not currently behave this way and would return an invalid datetime. It can be determined that this datetime is invalid by converting to UTC and back, which will return a different timestamp.

src/apscheduler/triggers/cron/__init__.py
```diff
@@ -202,22 +209,27 @@ def _set_field_value(self, dateval, fieldnum, new_value):
...
+                max(now.astimezone(UTC), self.start_date.astimezone(UTC)).astimezone(
+                    self.timezone
+                )
+                if self.start_date
+                else now
+            )
 
         fieldnum = 0
         next_date = datetime_ceil(start_date).astimezone(self.timezone)
+        print("start_next", repr(start_date))
```
agronholm: Remove this debug code.
```suggestion
```

src/apscheduler/job.py
```diff
@@ -145,7 +153,7 @@ def _get_run_times(self, now):
         """
         run_times = []
         next_run_time = self.next_run_time
-        while next_run_time and next_run_time <= now:
+        while next_run_time and next_run_time.astimezone(UTC) <= now.astimezone(UTC):
```
agronholm: Why is this necessary? datetime comparisons take the fold attribute into account.

agronholm (commented):
I'm not convinced that these UTC conversions are the right solution either. I may have to get my hands dirty and really dig into this problem.

src/apscheduler/triggers/cron/__init__.py
```diff
@@ -16,8 +17,16 @@
...
     datetime_repr,
+    datetime_utc_add,
 )
 
+if sys.version_info < (3, 11):
+    from datetime import timezone
+
+    UTC = timezone.utc
+else:
+    from datetime import UTC
```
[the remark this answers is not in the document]
agronholm: I think the `UTC = timezone.utc` aliasing is fine, but it doesn't need to be conditional on the Python version.

PR #1099, opened 2026-02-20
Add missing return statement in if_not_unset validator

bysiber:
## Summary

`if_not_unset()` defines an inner `validate` function but never returns it, so all attrs fields using this validator have no validation at all.

## Problem

```python
def if_not_unset(validator: Callable[[Any, Any, Any], None]) -> None:
    def validate(instance: Any, attribute: Any, value: Any) -> None:
        if value is unset:
            return
        validator(instance, attribute, value)
    # missing: return validate
```

The function implicitly returns `None`, so attrs treats the field as having no validator. This affects `TaskDefaults.job_executor` and all `TaskParameters` fields — they accept any value without validation.

## Fix

Add `return validate` at the end of the function.

### DIFF
--- a/src/apscheduler/_validators.py
+++ b/src/apscheduler/_validators.py
@@ -23,12 +23,14 @@ def aware_datetime(instance: Any, attribute: Attribute, value: Any) -> None:
         raise ValueError(f"{attribute.name} must be a timezone aware datetime")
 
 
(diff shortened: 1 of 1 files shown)


agronholm (approved):
Good catch, thanks!

PR #1107, opened 2026-04-02
Ensure wakeup() is called even when add_job() raises an exception

### LINKED ISSUE #1092: The failure of add_job resulted in not calling call wakeup
### Things to check first

### Version

3.11.0

### What happened?

I have a task where calling sqlalchemy's add_job to execute SQL failed, which caused wakeup not to be called, and thus other tasks were not executed.

The _real_add_job method in apscheduler\schedulers\base.py is as follows. An error occurred when calling store.add_job(job).

Image

This then caused the wakeup not to be called, which in turn affected other normal tasks and prevented them from being triggered.

```python
    def _real_add_job(self, job, jobstore_alias, replace_existing):
        """
        :param Job job: the job to add
        :param bool replace_existing: ``True`` to use update_job() in case the job already exists
            in the store

        """
        # Fill in undefined values with defaults
        replacements = {}
        for key, value in self._job_defaults.items():
            if not hasattr(job, key):
                replacements[key] = value

        # Calculate the next run time if there is none defined
        if not hasattr(job, "next_run_time"):
            now = datetime.now(self.timezone)
            replacements["next_run_time"] = job.trigger.get_next_fire_time(None, now)

        # Apply any replacements
        job._modify(**replacements)

        # Add the job to the given job store
        store = self._lookup_jobstore(jobstore_alias)
        try:
            store.add_job(job)
        except ConflictingIdError:
            if replace_existing:
                store.update_job(job)
            else:
                raise

        # Mark the job as no longer pending
        job._jobstore_alias = jobstore_alias

        # Notify listeners that a new job has been added
        event = JobEvent(EVENT_JOB_ADDED, job.id, jobstore_alias)
        self._dispatch_event(event)

        self._logger.info('Added job "%s" to job store "%s"', job.name, jobstore_alias)

        # Notify the scheduler about the new job
        if self.state == STATE_RUNNING:
            self.wakeup()
```

Reproduce the exception thrown by the add method in sqlalchemy store.

nuglifeleoji:
## Problem (fixes #1092)

If the job store's `add_job()` call raises an unexpected exception (e.g. a SQLAlchemy error during a database write), `_real_add_job()` propagates the exception before reaching the `self.wakeup()` call at the bottom of the method.

```python
# From _real_add_job():
try:
    store.add_job(job)          # <-- raises here
except ConflictingIdError:
    ...

# ... mark job, dispatch event, log ...

if self.state == STATE_RUNNING:
    self.wakeup()               # <-- never reached on unexpected exception
```

This leaves the scheduler's event loop without a wakeup signal. Already-scheduled, unrelated jobs stop being triggered until the scheduler is manually woken up.

## Fix

Wrap the body of `_real_add_job` in a `try/finally` so that `wakeup()` is always called when the scheduler is `STATE_RUNNING`, regardless of whether `add_job()` succeeded or failed.

The job-added event dispatch and log message are left inside the `try` block and are therefore **only emitted on success** — the semantics of those operations are unchanged.

```python
try:
    try:
        store.add_job(job)
    except ConflictingIdError:
        if replace_existing:
            store.update_job(job)
        else:
            raise
    # mark job, dispatch event, log ...
finally:
    if self.state == STATE_RUNNING:
        self.wakeup()
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/apscheduler/schedulers/base.py
+++ b/src/apscheduler/schedulers/base.py
@@ -1073,25 +1073,28 @@ class BaseScheduler(metaclass=ABCMeta):
         # Add the job to the given job store
         store = self._lookup_jobstore(jobstore_alias)
         try:
-            store.add_job(job)
-        except ConflictingIdError:
-            if replace_existing:
-                store.update_job(job)
-            else:
-                raise
+            try:
+                store.add_job(job)
+            except ConflictingIdError:
+                if replace_existing:
+                    store.update_job(job)
+                else:
+                    raise
 
-        # Mark the job as no longer pending
-        job._jobstore_alias = jobstore_alias
+            # Mark the job as no longer pending
+            job._jobstore_alias = jobstore_alias
 
-        # Notify listeners that a new job has been added
-        event = JobEvent(EVENT_JOB_ADDED, job.id, jobstore_alias)
-        self._dispatch_event(event)
+            # Notify listeners that a new job has been added
+            event = JobEvent(EVENT_JOB_ADDED, job.id, jobstore_alias)
+            self._dispatch_event(event)
 
-        self._logger.info('Added job "%s" to job store "%s"', job.name, jobstore_alias)
-
-        # Notify the scheduler about the new job
-        if self.state == STATE_RUNNING:
-            self.wakeup()
+            self._logger.info('Added job "%s" to job store "%s"', job.name, jobstore_alias)
+        finally:
+            # Notify the scheduler about the new job. This is done in a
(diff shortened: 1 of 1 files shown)


agronholm:
Why should the scheduler be woken up if adding the job failed? The whole premise seems odd to me. If the job store raises an exception from `add_job()`, then there is either a catastrophic failure or a bug in the job store. Either way, it is not the scheduler's job to catch such errors.

agronholm:
Judging by the fact that you force-pushed a revert which makes this PR a no-op, I conclude that you were not serious about this.

PR #1117, opened 2026-06-22
Fix numeric day_of_week range with step being ignored in CronTrigger

Labib-Bin-Salam:
## Changes

A numeric `day_of_week` range that has a step, such as `"1-5/2"`, silently dropped the step and matched **every** day in the range instead of every *n*th one. For example, `day_of_week="1-5/2"` fired on every weekday (Mon–Fri) rather than just Mon, Wed and Fri:

```python

t = CronTrigger(day_of_week="1-5/2", hour=0, timezone=timezone.utc,
                start_time=datetime(2024, 1, 1, tzinfo=timezone.utc))
print([t.next().strftime("%a") for _ in range(6)])
# before: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Mon']   <- step ignored
# after:  ['Mon', 'Wed', 'Fri', 'Mon', 'Wed', 'Fri']
```

APScheduler 3.x honored these steps (`DayOfWeekField("day_of_week", "1-5/2")` kept `1-5/2`), so this is a 4.x regression.

### Root cause

`DayOfWeekField.append_expression()` rewrites numeric weekday expressions into textual ones in order to remap cron's numbering (where both `0` and `7` mean Sunday) onto the internal, Monday-based numbering. That rewrite only carried the range endpoints and discarded the `/step`, and textual weekday ranges (`mon-fri`) have no step support to fall back on — so the step was lost and the range matched all of its days.

### Fix

Because cron's weekday numbering isn't linear in the internal numbering (e.g. `0-6/2` is `{sun, tue, thu, sat}` = internal `{6, 1, 3, 5}`), a stepped range can't be represented as a single internal range. So when a numeric range carries a step, expand it into the individual weekdays it matches. Non-stepped expressions and `*/step` are unchanged.

## Checklist

- [x] You've added tests (in `tests/`) which would fail without your patch
- [x] You've updated the documentation / changelog (in `docs/versionhistory.rst`)
- [x] You've added a new changelog entry (in `docs/versionhistory.rst`)

The new regression test `test_weekday_numeric_range_with_step` fails on `master` (fires Tue 2020-01-07 with `day_of_week='mon-fri'`) and passes with this change. `ruff`, `ruff format --check` and `mypy` are clean on the changed files.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -14,7 +14,8 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed an issue where ``CronTrigger.next()`` returned a non-existing date on a DST change
   (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_; PR by @jonasitzmann)
 - Fixed a numeric ``day_of_week`` range with a step (e.g. ``"1-5/2"``) ignoring the step in
-  ``CronTrigger`` and matching every day in the range instead (PR by @Labib-Bin-Salam)
+  ``CronTrigger`` and matching every day in the range instead
+  (`#1117 <https://github.com/agronholm/apscheduler/pull/1117>`_; PR by @Labib-Bin-Salam)
 - Fixed jobs that were being run when the scheduler was gracefully stopped being left in
   an acquired state (`#946 <https://github.com/agronholm/apscheduler/issues/946>`_)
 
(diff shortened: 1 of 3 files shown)


Labib-Bin-Salam:
Heads-up on CI: the failing jobs are entirely the pre-existing `cbor` serializer errors (`cbor2.CBORDecodeError: error decoding semantic tag 4664` / `'bool' object has no attribute 'tag'`) that currently turn every serialization test red — they reproduce identically on `master` and are unrelated to this change. The new regression test passes under the `json`/`pickle` serializers (it is among the 814 passed); only its `[cbor]` parametrization fails, like every other test. `pre-commit.ci` (ruff/ruff-format/mypy) and the docs build are green.

agronholm:
What prompted this? I think this was deliberate as range expressions in weekday numbers have very little practical use.

PR #1118, opened 2026-06-22
Fix CBOR serializer incompatibility with cbor2 6.0+

Labib-Bin-Salam:
## Changes

The CBOR serializer is broken against **cbor2 >= 6.0** (current release): every `CBORSerializer.deserialize()` raises `DeserializationError`, which currently turns all CBOR-parametrized tests red.

```python
from apscheduler.serializers.cbor import CBORSerializer

s = CBORSerializer()
t = DateTrigger(run_time=datetime(2020, 1, 1, tzinfo=timezone.utc))
s.deserialize(s.serialize(t))   # cbor2 6.x: apscheduler.DeserializationError
```

There are two root causes, both stemming from cbor2 6.0:

1. **`tag_hook` signature change.** cbor2 < 6 called the tag hook as `(decoder, tag, shareable_index=None)`; cbor2 >= 6 calls it as `(tag, immutable)`. So `_tag_hook` received the `CBORTag` in the wrong position and `tag.tag` evaluated on a `bool`, raising `'bool' object has no attribute 'tag'` → reported as `error decoding semantic tag 4664`. Fixed by locating the `CBORTag` among the arguments, so **both** calling conventions work (the `cbor2 >= 5.0` floor is kept). Also returns the tag unchanged for unrecognized tags (the documented cbor2 contract).

2. **Immutable tag contents.** cbor2 >= 6 deserializes the contents of a tag as immutable objects, so a `list` stored in an object's state comes back as a `tuple`. `OrTrigger`/`AndTrigger` keep a mutable list of next fire times and assign into it in `next()`, which raised `'tuple' object does not support item assignment` once deserialization got that far. Fixed by having `list_converter` coerce tuples back to lists (it is used only for that field).

## Verification

Locally with cbor2 6.1.2: the full `tests/triggers/` (164), `tests/test_serializers.py` (16) and `tests/test_marshalling.py` (14) suites pass; before this change every `[cbor]` parametrization in them failed. `ruff check`, `ruff format --check` and `mypy` are clean on the changed files.

Added a regression test (`test_serialize_trigger_with_mutable_list_state`) that round-trips an `OrTrigger` through the serializer and keeps using it — it covers both fixes and fails on `master` under `[cbor]`.

> Note: the remaining red jobs in CI are the pre-existing `test_datastores.py [asyncpg]`/`[mongodb]` **errors** (external services unavailable in the run) — they use the pickle serializer, are unrelated to CBOR, and are already failing on `master`.

## Checklist

- [x] You've added tests (in `tests/`) which would fail without your patch
- [x] You've updated the documentation / changelog (in `docs/versionhistory.rst`)
- [x] You've added a new changelog entry (in `docs/versionhistory.rst`)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -17,7 +17,8 @@ APScheduler, see the :doc:`migration section <migration>`.
   an acquired state (`#946 <https://github.com/agronholm/apscheduler/issues/946>`_)
 - Fixed the CBOR serializer raising ``DeserializationError`` with cbor2 v6.0 and newer,
   caused by a change in the ``tag_hook`` signature and tag contents now being deserialized
-  as immutable objects (PR by @Labib-Bin-Salam)
+  as immutable objects
+  (`#1118 <https://github.com/agronholm/apscheduler/pull/1118>`_; PR by @Labib-Bin-Salam)
 
 **4.0.0a6**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -15,6 +15,9 @@ APScheduler, see the :doc:`migration section <migration>`.
   (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_; PR by @jonasitzmann)
 - Fixed jobs that were being run when the scheduler was gracefully stopped being left in
   an acquired state (`#946 <https://github.com/agronholm/apscheduler/issues/946>`_)
+- Fixed the CBOR serializer raising ``DeserializationError`` with cbor2 v6.0 and newer,
+  caused by a change in the ``tag_hook`` signature and tag contents now being deserialized
+  as immutable objects (PR by @Labib-Bin-Salam)
 
 **4.0.0a6**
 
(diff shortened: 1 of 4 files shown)


Labib-Bin-Salam:
CI impact (test-linux 3.13): this PR takes the suite from **43 failed / 188 passed / 622 errors** on `master` to **0 failed / 234 passed / 622 errors** — every CBOR (de)serialization test now passes. The 622 `errors` are unchanged: they are the pre-existing `test_datastores.py [asyncpg]`/`[mongodb]` errors from the Postgres/MongoDB services not being reachable in the run (pickle, not CBOR), so they're unrelated to this change and also red on `master`.

agronholm:
Yet again the AI wildly hallucinates. The errors have nothing to do with MongoDB connectivity. There's nothing in the test output even suggesting that.

agronholm:
I think I might postpone this to the 7.0 release of cbor2 which should fix the immutability issue.

agronholm:
Thanks, but I've gone ahead and bumped the minimum to 6.0.

PR #1121, opened 2026-07-03
Fix scheduler crash when a job is rerun after its lease expired

### LINKED ISSUE #1116: AsyncScheduler crashes (KeyError) when a job's lease expires and the job is re-acquired/run twice (MemoryDataStore)
### Summary

With the in-memory data store, if a running job's lease expires before it
finishes, the scheduler re-acquires the same job and runs it a **second time**.
The duplicate run then crashes the scheduler on two unguarded operations:

- `MemoryDataStore.release_job`: `self._jobs_by_id.pop(result.job_id)` → `KeyError`
- `AsyncScheduler._run_job`: `self._running_jobs.remove(job)` → `KeyError`

The exception escapes the scheduler's TaskGroup, so `run_until_stopped()`
raises and the scheduler stops. In a long-lived service this is fatal.

Lease expiry of a still-running job is reachable in normal operation: the
lease-renewal coroutine (`extend_job_leases`, every `lease_duration / 2`) is
just another task on the event loop, so if the loop is saturated — e.g. a
burst of jobs that do blocking work inline at startup — the renewal is starved
and a concurrently-suspended job loses its lease.

### Reproduction (deterministic, ~5s)

```python
import asyncio, time
from apscheduler import AsyncScheduler

async def suspended_job():       # awaits across the lease window
    await asyncio.sleep(6)

async def loop_blocker():        # starves the lease-renewal coroutine
    time.sleep(6)

async def main():
    # short lease for a fast demo; in the wild it's a normal lease whose
    # renewal is starved by a saturated event loop
    async with AsyncScheduler(lease_duration=2, max_concurrent_jobs=5) as sched:
        now = datetime.now(timezone.utc)
        await sched.add_schedule(suspended_job, DateTrigger(run_time=now), id="job")
        await sched.add_schedule(loop_blocker, DateTrigger(run_time=now), id="blk")
        await sched.run_until_stopped()

asyncio.run(main())
```

### Expected

A job whose lease expired while it is still running should not be re-acquired
and run a second time; at minimum `release_job` / `_running_jobs.remove` should
tolerate a job that is already gone, instead of raising `KeyError` out of the
scheduler.

### Actual

```
Scheduler crashed
  ...
  File ".../apscheduler/datastores/memory.py", line 300, in release_job
    job = self._jobs_by_id.pop(result.job_id)
KeyError: UUID('...')
  ...
  File ".../apscheduler/_schedulers/async_.py", line 1238, in _run_job
    self._running_jobs.remove(job)
KeyError: Job(id=UUID('...'), task_id='__main__:suspended_job', ...)
```

### Notes

- The SQL and MongoDB stores release via an idempotent `DELETE`, so they don't
  crash on a redundant release; only `MemoryDataStore`'s `dict.pop` does. The
  `_running_jobs.remove` in `_run_job` is store-independent, though.
- Whether the *real* bug is "a still-running job should never be re-acquired"
  (the lease/renewal robustness) or "release must be idempotent" (defensive),
  the two unguarded `KeyError`s turn it into a hard scheduler crash either way.

### Environment

- apscheduler 4.0.0a6
- Python 3.14
- `MemoryDataStore` (the default for `AsyncScheduler()`)

leobaray:
## Changes

Fixes #1116.

When a job's lease expires while the job is still running (reachable in normal
operation: the lease renewal task is just another coroutine, so a saturated
event loop can starve it), `acquire_jobs()` hands the same job out again and the
scheduler runs a second copy. Since `Job` hashes/compares by `id`, the second
copy deduplicates into the single `_running_jobs` entry, and whichever run
finishes last crashes the scheduler with `KeyError` in two places
(`MemoryDataStore.release_job()` and `AsyncScheduler._run_job()`). The redundant
release also decremented the task's `running_jobs` counter a second time,
silently corrupting the slot accounting.

This PR takes both directions discussed in the issue, since they guard
different layers:

- **`AsyncScheduler`**: acquired jobs that are still running in this scheduler
  are skipped (with a warning) instead of being run a second time, so a lease
  expiry can no longer start a duplicate run in-process. Cross-scheduler
  duplicates remain possible by design (that's what leases are for), which is
  why the defensive fixes below still matter.
- **`AsyncScheduler._run_job()`**: unregisters the job with `discard()` instead
  of `remove()`, tolerating a duplicate run that was already unregistered.
- **`MemoryDataStore.release_job()`**: releasing an already-released job is now
  a no-op — like the idempotent `DELETE` the external data stores already do —
  and in particular no longer decrements the task's running jobs counter a
  second time.

The new test (`test_release_job_after_lease_expired`) exercises the
lease-expiry/steal/double-release sequence through the public data store API
and fails on the unpatched `MemoryDataStore` with the exact `KeyError` from the
issue. I ran it against the memory, sqlite and MySQL stores locally (the
PostgreSQL/MongoDB fixtures aren't runnable on my machine — ports taken by
local services); the issue's reproduction script now runs past the lease expiry
with the new warning logged instead of crashing.

## Checklist

features) — not applicable (bugfix; no documented behavior change)

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -15,6 +15,12 @@ APScheduler, see the :doc:`migration section <migration>`.
   (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_; PR by @jonasitzmann)
 - Fixed jobs that were being run when the scheduler was gracefully stopped being left in
   an acquired state (`#946 <https://github.com/agronholm/apscheduler/issues/946>`_)
+- Fixed the scheduler crashing with ``KeyError`` when a job's lease expired while the
+  job was still running and the job was then acquired (and run) a second time: the
+  scheduler now skips acquired jobs that are still running in it, and releasing an
+  already-released job is a no-op in ``MemoryDataStore`` (as it already was in the
+  external data stores), no longer corrupting the task's running jobs counter
+  (`#1116 <https://github.com/agronholm/apscheduler/issues/1116>`_; PR by @leobaray)
 
 **4.0.0a6**
 
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -1158,6 +1158,20 @@ class AsyncScheduler:
                         self.identity, self.lease_duration, limit
                     )
                     for job in jobs:
+                        # A job can be reacquired while still running in this
+                        # scheduler if its lease expired in the meantime (e.g. the
+                        # lease renewal task was starved by a saturated event
+                        # loop). Running a second copy would make the duplicate
+                        # release crash the scheduler (#1116).
+                        if job in self._running_jobs:
+                            self.logger.warning(
+                                "Skipping job %s: it was reacquired after its "
+                                "lease expired, but is still running in this "
+                                "scheduler",
+                                job.id,
+                            )
+                            continue
+
                         task = await self.data_store.get_task(job.task_id)
                         func = self._get_task_callable(task)
                         self._running_jobs.add(job)
@@ -1230,4 +1244,7 @@ class AsyncScheduler:
             finally:
                 current_job.reset(token)
         finally:
-            self._running_jobs.remove(job)
+            # discard() instead of remove(): if the job's lease expired mid-run and
+            # it was run a second time, the earlier finisher already removed the
+            # (id-equal) entry, and remove() would crash the scheduler (#1116)
+            self._running_jobs.discard(job)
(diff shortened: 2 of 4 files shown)


leobaray:
A note on the red CI: the failures appear to be pre-existing on master rather than caused by this change. The [last master run](https://github.com/agronholm/apscheduler/actions/runs/27916454159) fails with the same `cbor`-parametrized trigger tests (`tests/triggers/test_calendarinterval.py` etc.) and the same `test_datastores.py` fixture-setup cascade — starting at `test_add_replace_task[asyncmy]` at 0%, before anything from this PR runs (the `assert not self._finalizers` errors afterwards are pytest fixture-state fallout from that first failure).

For what it's worth, `test_release_job_after_lease_expired` passed against all eight data store fixtures in the jobs before that cascade point.

PR #1122, opened 2026-07-05
Preserve the callable when configuring a task from a Task object (#1001)

### LINKED ISSUE #1001: Method configure_task(Task(some_func)) doesn't register the callable in the scheduler (left at "None", causing the scheduler to crash)
### Things to check first

### Version

4.0.0a5post80

### What happened?

We're using version 4.0.0a5.post80 of apscheduler and discovered an issue when configuring tasks on a scheduler passing a "Task" object. Basically, the Callable I'm passing is not stored as part of the Scheduler Task.

```
import asyncio

from apscheduler import AsyncScheduler, Task

async def my_func():
    print("Hello World")
    return 42

async def main():
    async with scheduler as s:
        # Generate the Task object
        the_task = Task(id="test", func="__main__:my_func", job_executor="threadpool")
        # Register the Task with the Scheduler
        await s.configure_task(the_task)
        # print the tasks
        # Note that this line returns something like this:
        # [Task(id='test', func=None, job_executor='threadpool', max_running_jobs=None, misfire_grace_time=None, metadata={}, running_jobs=0)]
        print(await s.get_tasks())
        await s.add_job("test")

asyncio.run(main())
```

```
Scheduler crashed
  |   File "/workspaces/myproject/.venv/lib/python3.12/site-packages/apscheduler/_schedulers/async_.py", line 784, in run_until_stopped
  |                ^^^^^^^^^^^^^^^^^^^
  |   File "/workspaces/myproject/.venv/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 763, in __aexit__
  | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
    |   File "/workspaces/myproject/.venv/lib/python3.12/site-packages/apscheduler/_schedulers/async_.py", line 1012, in _process_jobs
    |     func = self._get_task_callable(task)
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/workspaces/myproject/.venv/lib/python3.12/site-packages/apscheduler/_schedulers/async_.py", line 974, in _get_task_callable
    |     raise CallableLookupError(
    | apscheduler._exceptions.CallableLookupError: Task test requires a locally defined callable to be run, but no such callable has been defined. Call scheduler.configure_task('test', func=...) to define the local callable.
    | 
  ...
```

When debugging, it turns out that the parameter passed [here](https://github.com/agronholm/apscheduler/blob/master/src/apscheduler/_schedulers/async_.py#L410) is actually None. This only happens as you pass a Task object to the scheduler. If you pass in the callable function directly, things work like a charm.

Your guidance / support is highly appreciated. Thanks a lot.

see above :)

uttam12331:
## Summary

Closes #1001.

`AsyncScheduler.configure_task()` accepts a callable, a task ID or a `Task` object. When passed a `Task`, it stored the task with `func=None`, so a later `add_job()` for that task crashed the scheduler with a callable lookup error:

```python
the_task = Task(id="test", func="__main__:my_func", job_executor="threadpool")
await scheduler.configure_task(the_task)
await scheduler.get_tasks()
# [Task(id='test', func=None, ...)]   <-- func was dropped
```

## Root cause

The `Task` branch of `configure_task()` built `TaskParameters` from the task's `id`, `job_executor`, `max_running_jobs`, `misfire_grace_time` and `metadata`, but never read `func_or_task_id.func`. The callable branch sets `func`, and the string branch reloads the stored task, but the `Task` branch left `func_ref` at its initial `None`, which is then used to build the task.

## Fix

Copy the task's callable reference into `func_ref` in the `Task` branch. Since `Task.func` is already a `"module:function"` reference string, it flows straight through to the created task. An explicitly passed `func=` keyword argument still takes precedence (it overrides `func_ref` via `callable_to_ref` further down), so no other behavior changes.

## Tests

Added `test_configure_task_from_task_object`, which configures a task from a `Task` object and asserts the stored `func` (and `job_executor`) are preserved. It fails on `master` (`func=None`) and passes with this change. `ruff check`/`ruff format` are clean; added a changelog entry under **UNRELEASED**.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -15,6 +15,9 @@ APScheduler, see the :doc:`migration section <migration>`.
   (`#1059 <https://github.com/agronholm/apscheduler/issues/1059>`_; PR by @jonasitzmann)
 - Fixed jobs that were being run when the scheduler was gracefully stopped being left in
   an acquired state (`#946 <https://github.com/agronholm/apscheduler/issues/946>`_)
+- Fixed ``configure_task()`` dropping the callable reference when passed a ``Task``
+  object, causing the configured task to be stored with ``func=None``
+  (`#1001 <https://github.com/agronholm/apscheduler/issues/1001>`_)
 
 **4.0.0a6**
 
--- a/src/apscheduler/_schedulers/async_.py
+++ b/src/apscheduler/_schedulers/async_.py
@@ -358,6 +358,10 @@ class AsyncScheduler:
             if func is unset:
                 func = func_or_task_id
         elif isinstance(func_or_task_id, Task):
+            # Carry over the callable reference stored on the task, otherwise the
+            # configured task would end up with func=None and the scheduler would
+            # fail to look up the callable when the job runs (#1001).
+            func_ref = func_or_task_id.func
             task_params = TaskParameters(
                 id=func_or_task_id.id,
                 job_executor=func_or_task_id.job_executor,
--- a/tests/test_schedulers.py
+++ b/tests/test_schedulers.py
@@ -47,6 +47,7 @@ from apscheduler import (
     SchedulerStarted,
     SchedulerStopped,
     ScheduleUpdated,
+    Task,
     TaskAdded,
     TaskDefaults,
     TaskUpdated,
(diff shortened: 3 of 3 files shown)


uttam12331:
Heads up: the failing CI jobs are unrelated to this change. The errors are all in `tests/test_datastores.py` (across every backend, including memory/aiosqlite and PyPy) and come from pytest's fixture teardown — `assert not self._finalizers` in `_pytest/fixtures.py` — i.e. a pytest / pytest-asyncio fixture-finalizer incompatibility, not anything in `test_schedulers.py`. This change only touches `configure_task` in `_schedulers/async_.py` and adds a scheduler test, both of which pass locally. Happy to rebase once the datastore-fixture issue is sorted on `master`.

PR #1127, opened 2026-07-25
Clarify numeric `day_of_week` cron convention in `CronTrigger`

### LINKED ISSUE #931: Improved documentation around "day_of_week" 0/1 confusion
### Things to check first

- [X] I have searched the existing issues and didn't find my feature already requested there

### Feature description

I think the `day_of_week` field mapping between days and integers in the `CronTrigger` class is confusing and violates the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). The `CronTrigger` is clearly named after the `cron` Unix tool, which, by convention, uses the integer `0` to refer to Sunday - however, the `apscheduler` implementation uses `0` as Monday. This nuance is not very clearly documented and this causes confusion to people migrating from e.g. `cron`.

While the decision to use `0` as Monday is not _strictly_ a bug in itself (although given the choice, I would have taken a different approach when implementing the class), I personally found the documentation to be unclear and the subsequent behaviour to be confusing.  I had to resort to reading the code to validate the behaviour when I was debugging a trigger.

I plan to add a PR to improve the docs if I get some time.

### Use case

Having clearer documentation of potential "expectation violations" will help reduce the friction experienced by developers when using the `CronTrigger` class.

cognis-digital:
## Changes

Fixes #931.

The `CronTrigger` docstring only stated that "the first weekday is always monday",
which conflicts with how numeric `day_of_week` input is actually interpreted on the
4.x branch: numbers follow the standard cron convention, where `0` and `7` both mean
Sunday, `1` means Monday and `6` means Saturday (the Monday-based ordering applies to
weekday *names* and name ranges, not to numeric values). Since `docs/api.rst`
autodocuments this class, the ambiguity is live in the published API reference — which
is exactly the confusion reported in #931.

This PR:
- Rewrites the `day_of_week` parameter description and the accompanying note to
  distinguish the numeric cron convention from the name-based (Monday-first) ordering,
  with a wrapping-range example.
- Fixes a small typo in the `day` parameter description ("day of the" -> "day of the
  month").
- Adds a parametrized regression test that pins the numeric-to-weekday mapping (0-7),
  so the documented cron convention can't be broken silently by future refactors.

## Checklist

- [x] Tests added (`tests/triggers/test_cron.py::test_numeric_day_of_week_follows_cron_convention`)
- [x] Documentation updated (`CronTrigger` docstring, which renders into `docs/api.rst`)
- [x] Changelog entry added (`docs/versionhistory.rst`, UNRELEASED)

### DIFF
--- a/src/apscheduler/triggers/cron/__init__.py
+++ b/src/apscheduler/triggers/cron/__init__.py
@@ -29,10 +29,11 @@ class CronTrigger(Trigger):
 
     :param year: 4-digit year
     :param month: month (1-12)
-    :param day: day of the (1-31)
+    :param day: day of the month (1-31)
     :param week: ISO week (1-53)
-    :param day_of_week: number or name of weekday (0-7 or sun,mon,tue,wed,thu,fri,sat,
-        sun)
+    :param day_of_week: number or name of weekday (0-7 or mon,tue,wed,thu,fri,sat,sun);
+        numeric values follow the standard cron convention where both ``0`` and ``7``
+        mean Sunday, ``1`` means Monday and ``6`` means Saturday
     :param hour: hour (0-23)
     :param minute: minute (0-59)
     :param second: second (0-59)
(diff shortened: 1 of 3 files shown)


src/apscheduler/triggers/cron/__init__.py
```diff
@@ -42,7 +43,11 @@ class CronTrigger(Trigger):
     :param timezone: time zone to use for the date/time calculations
         (defaults to the local timezone)
 
-    .. note:: The first weekday is always **monday**.
+    .. note:: Numeric ``day_of_week`` values follow the standard cron convention, in
+        which ``0`` and ``7`` both mean Sunday, ``1`` means Monday and so on up to ``6``
+        for Saturday. Weekday **names** and name ranges (such as ``mon-fri``) instead use
+        an ordering that starts on Monday, so a wrapping range like ``sat-tue`` covers
+        Saturday, Sunday, Monday and Tuesday.
```
agronholm: This seems like a waste of space. Unless there is widespread confusion about how the ranges wrap around, I would leave this out.
```suggestion
```

Issue #1053, opened 2025-06-04
wakeup_deadline will ignore all acquired schedules -> will not wake up on next fire time of a schedule currently acquired by another scheduler


Henkhogan:
### Things to check first

- [x] I have checked that my issue does not already have a solution in the [FAQ](https://apscheduler.readthedocs.io/en/master/faq.html)

- [x] I have searched the existing issues and didn't find my bug already reported there

- [x] I have checked that my bug is still present in the latest release

### Version

4.0.0a6

### What happened?

Whenever I start the scheduler and a certain schedule is still acquired (e.g. when I killed the last instance) it will be ignored. 

In my scenario I have job A running every minute and job B running daily. If job A is acquired, the scheduler will sleep until job B's firetime is reached.

"id"	"task_id"	"trigger"	"args"	"kwargs"	"paused"	"coalesce"	"misfire_grace_time"	"max_jitter"	"job_executor"	"job_result_expiration_time"	"metadata"	"last_fire_time"	"next_fire_time"	"acquired_by"	"acquired_until"
"exchange_scheduled_events_task"	"predictive.service.service:PredictiveService._exchange_scheduled_events_task"	"binary data"	"binary data"	"binary data"	false	"latest"			"async"	"00:00:00"	"{}"		"2025-06-05 00:05:00+00"		
"scheduled_events_task"	"predictive.service.service:PredictiveService._scheduled_events_task"	"binary data"	"binary data"	"binary data"	false	"latest"			"async"	"00:00:00"	"{}"		"2025-06-04 19:59:00+00"	"4edd430aa44d-1-548422838848"	"2025-06-04 19:59:30.002806+00"

The problem originates in the data stores get_next_schedule_run_time() method. For the SQLAlchemyDataStore it can be prevented by commenting out the filter on `acquired_by`

https://github.com/agronholm/apscheduler/blob/3c789566dc65161f77d7ebe1fe58f4b8b6064ecd/src/apscheduler/datastores/sqlalchemy.py#L826

### How can we reproduce the bug?

Mark a schedule as acquired in the DB by setting acquired_by and acquired_until. Then re-run the scheduler and see it ignored

agronholm:
> Ok. But since I have to support a distributed system with multiple schedulers, using a static ID will not work for me, right?

You should ideally assign each scheduler its own unique static ID if you can.

> Are you confirming the reported behaviour as a bug?

[...]

PR #1129, opened 2026-08-08
Release schedules with expired leases during clean-up

teddytennant:
Refs #1053.

Reworked to do it in housekeeping, as you suggested. `cleanup()` in all three data stores now
clears `acquired_by`/`acquired_until` on schedules whose leases have expired and publishes a
`ScheduleUpdated` for each one. A held schedule's `next_fire_time` is always in the past, so
`_process_schedules` wakes on the event and re-acquires it. The step is documented in the
`DataStore.cleanup()` contract in `abc.py`, next to the job lease one.

`get_next_schedule_run_time()` is back to exactly what master has.

Worth naming: recovery is now bounded by `cleanup_interval`, 15 minutes by default, rather than by
the lease duration. That is still a fix for a schedule that was otherwise stuck forever, and
`cleanup_interval` is the knob, but say the word if you want something tighter.

### Verification

Rebased onto current master first, so the cbor2 failures from the old run are gone.

```
pytest tests/test_datastores.py -k "memory or aiosqlite or mongodb"
73 passed, 1 skipped
```

`test_next_schedule_run_time_acquired` is replaced by `test_cleanup_expired_schedule_leases`, which
acquires a schedule, checks clean-up leaves a live lease alone, then checks it releases the schedule
and emits `ScheduleUpdated` once the lease expires. With `src/apscheduler/datastores/` reverted to
master it fails on all three of memory, aiosqlite and mongodb.

PostgreSQL and MySQL I still can't run locally. `ruff check` and `ruff format --check` are unchanged
from master.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -19,6 +19,10 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed the documented ``psycopg`` extra not being declared in the project metadata, so
   ``pip install apscheduler[psycopg]`` silently installed nothing for the Psycopg event
   broker (`#1133 <https://github.com/agronholm/apscheduler/issues/1133>`_)
+- Fixed schedules staying stuck when the scheduler holding them died without releasing
+  them; clean-up now releases schedules whose leases have expired and wakes up the
+  other schedulers so they can acquire them
+  (`#1053 <https://github.com/agronholm/apscheduler/issues/1053>`_)
 
 **4.0.0a6**
 
--- a/src/apscheduler/abc.py
+++ b/src/apscheduler/abc.py
@@ -402,6 +402,8 @@ class DataStore(metaclass=ABCMeta):
         * Purge expired job results (where ``expires_at`` is less or equal to the
           current time)
         * Release jobs with expired leases with the ``cancelled`` outcome
+        * Release schedules with expired leases, publishing a
+          :class:`~apscheduler.ScheduleUpdated` event for each of them
         * Purge finished schedules (where ``next_run_time`` is ``None``) that have no
           running jobs associated with them
         """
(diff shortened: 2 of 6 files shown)


teddytennant:
These failures are pre-existing on master, not caused by this PR.

Every failing job fails on the 43 `[cbor]` parametrizations, and they fail identically on this branch's merge base (26bff5d). That commit's own master run, [29191123853](https://github.com/agronholm/apscheduler/actions/runs/29191123853), is red on all 10 jobs with 43 failures each — 44 on test-linux 3.10, the extra one being the flaky `test_scheduler_crash_reap_abandoned_jobs[asyncpg]`. Side by side on 3.10: master 44 failed / 811 passed, this PR 44 failed / 819 passed (same failures, 8 more passing tests).

The cause is cbor2 6.0.0, which changed the tag hook calling convention from `(decoder, tag)` to `(tag, immutable)`:

```python
# cbor2/__init__.pyi (6.1.4)
TagHook: TypeAlias = Callable[[CBORTag, bool], Any]
```

So `CBORSerializer._tag_hook` receives the `immutable` bool in the position where it expects the `CBORTag`, giving `AttributeError: 'bool' object has no attribute 'tag'`, which surfaces as `CBORDecodeError: error decoding semantic tag 4664`. The `cbor` extra allows `cbor2 >= 5.0`, so CI resolves 6.1.4. Locally on this branch, `pytest tests/triggers`: 164 passed with cbor2 5.9.0, 43 failed with 6.1.4.

#1118 already fixes this, including a second cbor2 6.x behaviour change (tag contents now decode as immutable, so `OrTrigger._next_fire_times` comes back as a tuple and `next()` raises `TypeError: 'tuple' object does not support item assignment`). I've deliberately kept that out of this PR so the diff stays scoped to the wakeup deadline — happy to rebase once it lands.

agronholm:
Seems to me like housekeeping should free up expired leases for schedules. If that happens, the schedulers should be woken up.

teddytennant:
Good call, moved it into `cleanup()`. All three stores now release schedules whose leases have
expired and publish `ScheduleUpdated`, and `get_next_schedule_run_time()` is back to what master
has. Also rebased, the old CI red was the cbor2 breakage you've since fixed.

agronholm (approved):
Solid work, thanks!

PR #1130, opened 2026-08-13
Raise ValueError for out-of-range numeric day_of_week values

nikolauspschuetz:
## Changes

No existing issue — linking to this PR in the changelog.

`CronTrigger` accepts numeric `day_of_week` values in the documented `0-7` range
(where both `0` and `7` mean Sunday). A value outside that range currently raises a
bare `IndexError: list index out of range` instead of a helpful error, because the
numeric value is mapped straight into `WEEKDAYS` before any range validation runs:

- `CronTrigger(day_of_week="8")` → `IndexError`
- `CronTrigger(day_of_week="0-8")` → `IndexError` (the range endpoint)
- `CronTrigger(day_of_week="7")` → OK (Sunday)

This validates the numeric bounds before the mapping and raises a descriptive
`ValueError`, mirroring the existing `Invalid weekday name '...'` messages in the same
module (e.g. `ValueError: Invalid weekday number '8'`).

## Checklist

- [ ] You've updated the documentation (in `docs/`) — n/a, error-message behavior only

### Notes

- The regression test `test_invalid_weekday_number` (parametrized `"8"`, `"0-8"`) fails on
  `master` with `IndexError` and passes with this patch.
- This only touches the numeric weekday-range branch (`first`/`last`). It does **not** touch
  the step branch (`groups[2]`, e.g. `"1-5/2"`) addressed by #1117 — no functional overlap,
  though the two changes are textually adjacent in the same method.

<sub>drafted with claude code — just for the record</sub>

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -26,6 +26,9 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed the synchronous ``Scheduler`` not enabling automatic clean-up by default,
   unlike ``AsyncScheduler``
   (`#1135 <https://github.com/agronholm/apscheduler/pull/1135>`_; PR by @dylanpulver)
+- Fixed ``CronTrigger`` raising an unhelpful ``IndexError`` instead of a descriptive
+  ``ValueError`` when given a numeric ``day_of_week`` value outside the 0–7 range
+  (`#1130 <https://github.com/agronholm/apscheduler/pull/1130>`_; PR by @nikolauspschuetz)
 
 **4.0.0a6**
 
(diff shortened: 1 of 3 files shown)


nikolauspschuetz:
Heads-up on the red CI — it isn't this change. Every `[cbor]` parametrization fails with `AttributeError: 'bool' object has no attribute 'tag'` at `serializers/cbor.py:56`; a recent cbor2 release changed the `tag_hook` signature. The `[pickle]`/`[json]` variants of the same tests pass, and the new `test_invalid_*` cases here are green. Nothing in this PR touches the serializers, so it won't go green until cbor2 is pinned/fixed on main.

agronholm (approved):
LGTM, thanks!

PR #1132, opened 2026-08-17
Keep the wakeup timer rearmed if _process_jobs raises exception.

sundasnoreen-sr:
## Changes

Fixes (no existing issue)

### The problem

`AsyncIOScheduler.wakeup()` did three things in order:

1. stop the current timer
2. call `_process_jobs()`, which returns how long to sleep
3. start a new timer using that value

If step 2 raised an exception, step 3 never ran. The timer was already stopped
in step 1, so at that point **nothing was scheduled to wake the scheduler up
again**. `wakeup()` is invoked through `call_soon_threadsafe`, so the exception
just surfaced in the event loop and was lost. The scheduler stayed alive and
looked healthy, but it never ran another job.

The only way it could recover was by luck: any external call that happens to
invoke `wakeup()` again — `add_job()`, `modify_job()`, `resume()` — would start
a fresh timer. Nothing in the scheduler itself recovered on its own.

### What we observed in production

We hit this with `SQLAlchemyJobStore` on PostgreSQL:

1. `_process_jobs()` finished running the due jobs and called
   `jobstore.update_job()` to persist each job's next run time.
2. The PostgreSQL server closed the connection during the `COMMIT`.
3. The resulting exception escaped `_process_jobs()`, so the timer was never
   restarted.
4. The scheduler sat completely idle — no jobs ran, no errors were logged —
   until an unrelated `add_job()` call happened to re-arm the timer and bring
   it back to life.

A dropped database connection is a routine, transient event. It should not
permanently disable the scheduler, and it should not fail silently.

### The fix

Wrap the `_process_jobs()` call so that a failure is logged and the timer is
rearmed after `jobstore_retry_interval` instead of being left stopped. This
matches how `BlockingScheduler` already handles job store failures, and it
turns a permanent silent outage into a logged, self-healing retry.

## Checklist

- [ ] You've updated the documentation (in `docs/`) — n/a, bug fix only

### DIFF
--- a/tests/test_schedulers.py
+++ b/tests/test_schedulers.py
@@ -1281,17 +1281,17 @@ class TestAsyncIOScheduler(SchedulerImplementationTestBase):
         scheduler does not stop processing jobs entirely.
 
         """
-        caplog.set_level(logging.ERROR, logger="apscheduler.schedulers")
+        caplog.set_level(logging.ERROR, logger="apscheduler.scheduler")
+        scheduler.jobstore_retry_interval = 3
         scheduler._eventloop = MagicMock()
         scheduler._process_jobs = MagicMock(side_effect=RuntimeError("Boom"))
 
         # Bypass the run_in_event_loop decorator to run wakeup() synchronously
         AsyncIOScheduler.wakeup.__wrapped__(scheduler)
 
-        scheduler._eventloop.call_later.assert_called_once_with(
-            scheduler.jobstore_retry_interval, scheduler.wakeup
-        )
+        scheduler._eventloop.call_later.assert_called_once_with(3, scheduler.wakeup)
         assert "Error processing jobs" in caplog.text
+        assert "RuntimeError: Boom" in caplog.text
 
 
 class TestGeventScheduler(SchedulerImplementationTestBase):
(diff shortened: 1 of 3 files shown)


tests/test_schedulers.py
```diff
@@ -1275,6 +1275,24 @@ def start_scheduler(self, request, scheduler):
...
+        scheduler._eventloop = MagicMock()
+        scheduler._process_jobs = MagicMock(side_effect=RuntimeError("Boom"))
+
+        # Bypass the run_in_event_loop decorator to run wakeup() synchronously
+        AsyncIOScheduler.wakeup.__wrapped__(scheduler)
+
+        scheduler._eventloop.call_later.assert_called_once_with(
+            scheduler.jobstore_retry_interval, scheduler.wakeup
+        )
+        assert "Error processing jobs" in caplog.text
```
agronholm, line 1294: Add an assertion for "Boom" and for the rearming using the job store retry interval.

agronholm (changes requested):
Seems reasonable. I would like a couple of additions to the test before I merge.

tests/test_schedulers.py
sundasnoreen-sr: Updated.

agronholm (approved):
Thanks.

PR #1135, opened 2026-08-26
Fixed the sync scheduler never running automatic clean-up

dylanpulver:
## Changes

`Scheduler` is documented as a synchronous wrapper for `AsyncScheduler`, pointing at that
class for the meaning of its configuration options. One default does not match:

```python
>>> from apscheduler import AsyncScheduler, Scheduler
>>> AsyncScheduler().cleanup_interval
datetime.timedelta(seconds=900)
>>> Scheduler().cleanup_interval
None
```

`Scheduler.__init__` declares `cleanup_interval: float | timedelta | None = None` and
forwards the value unconditionally, so `AsyncScheduler` receives an explicit `None`.
`None` is the documented "disable automatic clean-up" value, so `_cleanup_loop` never
starts for a sync scheduler built with default options and the data store keeps expired
job results and finished schedules forever. The user guide states that each scheduler
runs the data store's `cleanup()` method periodically.

Both defaults arrived in 56274e4, which added the option to the two classes in one
commit and gave the sync side `None`. The other scalar options in that signature do
mirror the async defaults (`max_concurrent_jobs=100`,
`lease_duration=timedelta(seconds=30)`), and `test_interface_parity` skips dunder names,
so `__init__` is the one part of the wrapper the parity test never compares.

The fix mirrors the async default the way the sibling options do. One added test asserts
directly on `Scheduler().cleanup_interval` and on parity with `AsyncScheduler()`. The
other checks that a default sync scheduler reaches the data store's `cleanup()`. Both
fail on master.

## Checklist

features)

### DIFF
--- a/tests/test_schedulers.py
+++ b/tests/test_schedulers.py
@@ -1498,15 +1498,6 @@ class TestSyncScheduler:
             with pytest.raises(ScheduleLookupError):
                 scheduler.get_schedule("event_set")
 
-    def test_default_cleanup_interval(self) -> None:
-        """
-        Test that the sync scheduler defaults to the same clean-up interval as the
-        async scheduler instead of disabling automatic clean-up.
-
-        """
-        assert Scheduler().cleanup_interval == timedelta(minutes=15)
-        assert Scheduler().cleanup_interval == AsyncScheduler().cleanup_interval
-
     def test_implicit_cleanup(self, mocker: MockerFixture) -> None:
         """
         Test that the data store's cleanup() method is called when a scheduler
--- a/tests/test_schedulers.py
+++ b/tests/test_schedulers.py
@@ -16,6 +16,7 @@ from typing import Any, cast
 from unittest.mock import patch
 
 import anyio
+import attrs
 import pytest
 from anyio import (
     Lock,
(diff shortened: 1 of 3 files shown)


agronholm:
This feels like a gap in `test_interface_parity` which should have caught this, but I don't think we test parameter defaults there.

dylanpulver:
You're right, defaults weren't covered. The loop only walks public attributes, so `__init__` never gets compared at all.

Added `test_constructor_parity`. It walks `AsyncScheduler.__init__` and compares each argument's declared default against `Scheduler()`, running both through the attrs field converter first so that `lease_duration=30` and `timedelta(seconds=30)` come out equal. It also fails if `Scheduler()` is missing an argument outright.

Excluded: `data_store`, `event_broker`, `job_executors`, `task_defaults`, `logger`. `Scheduler` only forwards the options it was given and uses `None` as the "not passed" sentinel for anything `AsyncScheduler` builds itself, so the effective default still comes from `AsyncScheduler`. `task_defaults` diverges on purpose too, since `Scheduler` fills in the threadpool executor.

Confirmed the new test isn't vacuous by reverting the one-line fix: it fails with `Default value mismatch for the 'cleanup_interval' argument of Scheduler(): None != datetime.timedelta(seconds=900)`, and `test_interface_parity` still passes on its own, which is the gap you spotted.

tests/test_schedulers.py
```diff
@@ -1446,6 +1498,27 @@ def test_explicit_cleanup(self) -> None:
...
+        Test that the data store's cleanup() method is called when a scheduler
+        configured with the default options is started.
+
+        """
+        with Scheduler() as scheduler:
+            event = threading.Event()
+            mocker.patch.object(scheduler.data_store, "cleanup", side_effect=event.set)
+            scheduler.start_in_background()
+            assert event.wait(3)
+
```
agronholm: Seems to me that your newly added test makes these redundant.
```suggestion
```

dylanpulver:
Half of it, yes — the second assertion is now covered by `test_constructor_parity` and I'll drop it.

The first one isn't redundant though. `test_constructor_parity` only checks the two schedulers agree with *each other*, so a coordinated change sails past it. Setting both defaults to 20 minutes:

```
test_interface_parity      PASSED
test_constructor_parity    PASSED
test_default_cleanup_interval  FAILED
  assert Scheduler().cleanup_interval == timedelta(minutes=15)
  AssertionError: assert timedelta(seconds=1200) == timedelta(seconds=900)
```

It's also the only check on the instantiated attribute rather than the declared signature default.

So unless you'd rather see it go anyway, I'd keep line 1507 and delete just 1508. Happy to do either — say which and I'll push.

agronholm:
> Half of it, yes — the second assertion is now covered by `test_constructor_parity` and I'll drop it.
> 
> The first one isn't redundant though. `test_constructor_parity` only checks the two schedulers agree with _each other_, so a coordinated change sails past it. Setting both defaults to 20 minutes:
> 
> ```
> test_interface_parity      PASSED
> test_constructor_parity    PASSED
> test_default_cleanup_interval  FAILED
>   assert Scheduler().cleanup_interval == timedelta(minutes=15)
>   AssertionError: assert timedelta(seconds=1200) == timedelta(seconds=900)
> ```
> 
> It's also the only check on the instantiated attribute rather than the declared signature default.
> 
> So unless you'd rather see it go anyway, I'd keep line 1507 and delete just 1508. Happy to do either — say which and I'll push.

Testing for a specific value seems pointless. We don't have similar tests for other constructor values anyway. Please drop the test.

dylanpulver:
Dropped, in e4f2746. Reverting the `sync.py` change still fails `test_constructor_parity`, so the fix stays covered.

agronholm:
Thanks!

PR #1136, opened 2026-09-05
Fix AndTrigger.max_iterations type to match its actual behavior

### LINKED ISSUE #1131: max_iterations=None on AndTrigger raises TypeError even though the annotation allows None
`AndTrigger.max_iterations` is annotated `int | None` and defaults to 10000, but `next()` does `range(self.max_iterations)`, so passing `None` raises instead of doing anything useful.

```python
from apscheduler.triggers.combining import AndTrigger

t = AndTrigger([IntervalTrigger(seconds=1), IntervalTrigger(seconds=1)], max_iterations=None)
t.next()
# TypeError: 'NoneType' object cannot be interpreted as an integer
```

It constructs fine. The same trigger with the default returns a datetime.

Two ways to read it, and I do not want to guess which you want:

Treat `None` as no limit, which is what the annotation suggests and what `None` usually means for a cap. The cost is that a pair of triggers whose fire times never fall within `threshold` would loop until something else stops it, and the cap looks like it exists precisely to prevent that.

Or narrow the annotation to `int`, since the current behaviour is already "a number is required" and the anti-hang guarantee stays intact. That makes the type match the code and closes the escape hatch.

I lean toward the second because the cap reads as deliberate, but you know whether an unlimited mode was ever intended.

Happy to send a PR for whichever you prefer.

afonsojanu:
Closes #1131.

`max_iterations` is annotated `int | None`, but passing `None` just crashes inside `next()` with `TypeError: 'NoneType' object cannot be interpreted as an integer` the moment it hits `range(self.max_iterations)`. There's no code path anywhere that treats `None` as unlimited.

Looking at what the cap is actually for, it backs `MaxIterationsReached`, which exists specifically to stop a pair of triggers whose fire times never line up from looping forever. That's a deliberate guard, so making `None` mean unlimited would defeat the point of having it. Narrowing the type to `int` seemed like the smaller, safer change versus rewriting the loop to support an unbounded mode, which is also what the issue reporter leaned toward.

Also added a test that pins down the current TypeError-on-None behavior, so this doesn't get silently "fixed" into an infinite loop down the road.

Ran the full trigger test suite locally (165 tests, all passing) plus a manual repro of the original crash before/after the change.

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -26,6 +26,10 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed the synchronous ``Scheduler`` not enabling automatic clean-up by default,
   unlike ``AsyncScheduler``
   (`#1135 <https://github.com/agronholm/apscheduler/pull/1135>`_; PR by @dylanpulver)
+- Fixed ``AndTrigger.max_iterations`` being typed as ``int | None`` even though passing
+  ``None`` just raises a confusing ``TypeError`` from inside ``next()``. The
+  iteration cap is deliberate, so the type now says what's actually required
+  (`#1131 <https://github.com/agronholm/apscheduler/issues/1131>`_)
 
 **4.0.0a6**
 
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -26,9 +26,8 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed the synchronous ``Scheduler`` not enabling automatic clean-up by default,
   unlike ``AsyncScheduler``
   (`#1135 <https://github.com/agronholm/apscheduler/pull/1135>`_; PR by @dylanpulver)
-- Fixed ``AndTrigger.max_iterations`` being typed as ``int | None`` even though passing
-  ``None`` just raises a confusing ``TypeError`` from inside ``next()``. The
-  iteration cap is deliberate, so the type now says what's actually required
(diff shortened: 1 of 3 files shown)


docs/versionhistory.rst
```diff
@@ -26,6 +26,10 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed the synchronous ``Scheduler`` not enabling automatic clean-up by default,
   unlike ``AsyncScheduler``
   (`#1135 <https://github.com/agronholm/apscheduler/pull/1135>`_; PR by @dylanpulver)
+- Fixed ``AndTrigger.max_iterations`` being typed as ``int | None`` even though passing
+  ``None`` just raises a confusing ``TypeError`` from inside ``next()``. The
+  iteration cap is deliberate, so the type now says what's actually required
```
agronholm, line 31: Reduced verbosity
```suggestion
- Fixed type annotation of ``AndTrigger.max_iterations`` to only allow ``int`` and never
  ``None``
```

tests/triggers/test_combining.py
```diff
@@ -44,6 +44,22 @@ def test_max_iterations(self, timezone, serializer):
...
+                IntervalTrigger(seconds=4, start_time=start_time),
+                IntervalTrigger(
+                    seconds=4, start_time=start_time + timedelta(seconds=2)
+                ),
+            ],
+            max_iterations=None,
+        )
+
+        pytest.raises(TypeError, trigger.next)
+
```
agronholm, line 62: I don't think a negative test is warranted here.
```suggestion
```

PR #1137, opened 2026-09-09
Fix combining trigger comparisons across DST transitions

Kuang-xianxin:

## Changes

During the October clock rollback in Europe/Berlin, `02:30+02:00` and `02:30+01:00` are an hour apart. Python compares these datetimes as equal when they share the same `ZoneInfo`, so `OrTrigger` can lose a run, return rollback times out of order, or duplicate the same instant represented in another timezone. `AndTrigger` can accept times an hour apart within a one-second threshold and discard times only half a second apart across a DST transition.

Use `absolute_datetime_diff()` for comparisons, deduplication and thresholds. The helper subtracts UTC datetimes before converting the difference to seconds, preserving microsecond threshold boundaries that epoch-float subtraction can round incorrectly. The `OrTrigger` minimum uses a UTC datetime key. Original child datetimes and serialized state remain unchanged.

Validation refreshed on September 23, 2026 (Windows, CPython 3.13.14), after merging master `81e9fdcaac8d6c72f4318cb7ff393c913c7f9440` and retaining the maintainer's commits:

- Against unmodified current master source, the existing DST regressions still fail in all 22 cases; the microsecond control passes. The cases cover both child orders, both folds, spring-forward/fall-back thresholds and partially consumed trigger round-trips through pickle, CBOR and JSON.
- On the merged PR: `pytest -m 'not external_service' -q` passes **415 tests**, with 2 skipped and 492 deselected. External database/message-broker services were not run.
- Applicable automatic pre-commit hooks pass for the PR files and the files brought in by the merge; `git diff --check` passes. The merge resolves only the adjacent changelog entries; this PR's implementation and regression tests are unchanged.
- The existing microsecond regression was previously verified to reject the helper's epoch-float subtraction (an exact 1-microsecond interval becomes approximately 1.192 microseconds). UTC subtraction preserves the boundary.
- Optional manual mypy and strict Sphinx checks were not rerun for this changelog-only conflict resolution. The September 10 validation found baseline typing diagnostics and documentation type-reference warnings; no complete type-check or strict documentation-build pass is claimed.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -437,11 +437,6 @@ If this trigger is created on 2022-06-07 at 09:00:00, its first run times would 
 
 Notably, 2022-08-07 is skipped because it falls on a Sunday.
 
-Both combining triggers compare run times as absolute instants. During a daylight saving
-time rollback, two occurrences of the same local clock time remain distinct. The
-``AndTrigger`` threshold measures elapsed time, including across daylight saving time
-transitions. Returned run times retain the time zones of their originating triggers.
-
 Removing schedules
 ------------------
 
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -437,6 +437,11 @@ If this trigger is created on 2022-06-07 at 09:00:00, its first run times would 
 
 Notably, 2022-08-07 is skipped because it falls on a Sunday.
 
+Both combining triggers compare run times as absolute instants. During a daylight saving
+time rollback, two occurrences of the same local clock time remain distinct. The
+``AndTrigger`` threshold measures elapsed time, including across daylight saving time
+transitions. Returned run times retain the time zones of their originating triggers.
+
 Removing schedules
 ------------------
 
--- a/src/apscheduler/triggers/combining.py
+++ b/src/apscheduler/triggers/combining.py
@@ -9,7 +9,7 @@ import attrs
 from .._converters import as_aware_datetime, as_timedelta, list_converter
 from .._exceptions import MaxIterationsReached
 from .._marshalling import marshal_object, unmarshal_object
-from .._utils import require_state_version
+from .._utils import absolute_datetime_diff, require_state_version
 from ..abc import Trigger
 
 
@@ -65,6 +65,7 @@ class AndTrigger(BaseCombiningTrigger):
             # Fill out the fire times on the first run
             self._next_fire_times = [t.next() for t in self.triggers]
 
+        threshold = self.threshold.total_seconds()
         for _ in range(self.max_iterations):
             # Find the earliest and latest fire times
             earliest_fire_time: datetime | None = None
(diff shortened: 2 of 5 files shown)


docs/userguide.rst
```diff
@@ -437,6 +437,11 @@ If this trigger is created on 2022-06-07 at 09:00:00, its first run times would
 
 Notably, 2022-08-07 is skipped because it falls on a Sunday.
 
+Both combining triggers compare run times as absolute instants. During a daylight saving
+time rollback, two occurrences of the same local clock time remain distinct. The
+``AndTrigger`` threshold measures elapsed time, including across daylight saving time
+transitions. Returned run times retain the time zones of their originating triggers.
+
```
agronholm: This smells like an AI-generated vacuous paragraph. Why are these technical details necessary to add to the user guide?

src/apscheduler/triggers/combining.py
```diff
@@ -74,19 +74,33 @@ def next(self) -> datetime | None:
                 if fire_time is None:
                     return None
 
-                if earliest_fire_time is None or earliest_fire_time > fire_time:
+                # Compare instants, not wall times: datetime comparisons ignore fold
+                # when both operands share the same tzinfo.
+                if earliest_fire_time is None or earliest_fire_time.astimezone(
+                    timezone.utc
+                ) > fire_time.astimezone(timezone.utc):
```
agronholm: Consider using `absolute_datetime_diff()` from `apscheduler._utils`.

docs/userguide.rst
Kuang-xianxin: Removed the paragraph in 6fef770. Agreed: it repeated implementation details without adding useful user guidance. The changelog entry remains.

src/apscheduler/triggers/combining.py
Kuang-xianxin: Reused `absolute_datetime_diff()` in 6fef770 for ordering, deduplication and threshold checks. I also changed the helper to subtract UTC datetimes before converting the difference to seconds: its epoch-float subtraction returns approximately 1.192 microseconds for an exact 1-microsecond difference, causing `AndTrigger` to reject a valid 1-microsecond threshold. The added regression reproduces that failure and passes with UTC subtraction.

The non-external suite passes: 395 tests, 2 skipped. Applicable automatic pre-commit hooks pass. The optional manual mypy check still reports baseline typing issues (72 diagnostics on both versions; details in the PR body).

Kuang-xianxin:
Merged current master (`81e9fdc`) in `4a0461846fa9a2f9146ec8fdc991190b2dad0b9b`, preserving your `0774b4a` and `d011ab3` commits. The only conflict was between adjacent changelog entries; both this fix and the upstream CronTrigger entry are retained. The combining-trigger implementation and its tests are unchanged.

Revalidated on September 23 (Windows, CPython 3.13.14):

- The existing DST regressions still fail in all 22 cases against unchanged current master; the microsecond control passes.
- The merged PR passes **415 non-external-service tests**, with 2 skipped and 492 deselected (`pytest -m 'not external_service' -q`).
- Applicable automatic pre-commit checks and `git diff --check` pass. External database/message-broker services, optional manual mypy and strict Sphinx checks were not rerun.

The PR description now reflects this verification. Ready for re-review; the new head's CI results will be tracked separately from the previous head.

PR #1138, opened 2026-09-10
Fix calendar interval dates after midnight DST gaps

Yukibei:

## Changes

In `America/Nuuk`, a daily 23:30 `CalendarIntervalTrigger` starting on 2024-03-29 skips the valid March 31 run after the March 30 DST gap:

- Before: March 29 → April 1 → April 2.
- After: March 29 → March 31 → April 1.

The nonexistent March 30 23:30 normalizes to March 31 00:30. After rejecting this time, the original implementation uses the normalized date as the next iteration's reference, then adds the interval again. This can also shift subsequent weekly, monthly and yearly dates, or exhaust a trigger before its valid inclusive end date.

Keep the attempted `next_date` as the reference when discarding an invalid local time. The existing interval calculation, wall-clock time, end-date handling and serialized state format remain unchanged. No dependencies are added.

Validation on Windows, CPython 3.12.10, tzdata 2026.3, based on `e1afa7471d1fc2c3ea7a2b730ef68db20743774a`:

- All 18 new regression cases fail on the original implementation. They cover daily, every-other-day, weekly, monthly and yearly intervals, a start date inside the gap, the inclusive end date, and repeated pickle/CBOR/JSON round-trips.
- `pytest tests/triggers tests/test_serializers.py -q`: 197 passed, including the new cases.
- A separate public-API reproduction produces the expected March 29 → March 31 → April 1 sequence after the fix.
- All applicable automatic pre-commit hooks pass for the three changed files, and `git diff --check` passes.

The existing `CalendarIntervalTrigger` documentation already specifies discarding nonexistent local times and advancing to the next calendar date, so no user-guide changes are needed. The full cross-platform Python matrix, external-service suite and optional manual mypy hook were not run locally. The changelog entry will be linked to this PR once its number is assigned.

AI assistance: Codex assisted with investigation, implementation and regression tests. The validation above was executed locally against the modified source.

## Checklist

features)

these instructions.

### Updating the changelog

```
```

### DIFF
--- a/docs/versionhistory.rst
+++ b/docs/versionhistory.rst
@@ -29,6 +29,8 @@ APScheduler, see the :doc:`migration section <migration>`.
 - Fixed type annotation of ``AndTrigger.max_iterations`` to only allow ``int`` and never
   ``None``
   (`#1131 <https://github.com/agronholm/apscheduler/issues/1131>`_; PR by @afonsojanu)
+- Fixed ``CalendarIntervalTrigger`` skipping valid dates or shifting subsequent run dates
+  when a nonexistent local time normalizes past midnight during a forward DST shift
 
 **4.0.0a6**
 
--- a/src/apscheduler/triggers/calendarinterval.py
+++ b/src/apscheduler/triggers/calendarinterval.py
@@ -119,7 +119,7 @@ class CalendarIntervalTrigger(Trigger):
 
             # Check if the time is off due to normalization and a forward DST shift
             if next_time.timetz() != self._time:
-                previous_date = next_time.date()
+                previous_date = next_date
             else:
                 self._last_fire_date = next_date
                 return next_time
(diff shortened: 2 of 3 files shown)


agronholm:
The documentation explicitly warns about picking times:
```rst
    If the resulting datetime is invalid due to a daylight saving forward shift, the
    date is discarded and the process moves on to the next date. If instead the datetime
    is ambiguous due to a backward DST shift, the earlier of the two resulting datetimes
    is used.
```
The promise here is that the selected wall-clock time is respected. Your PR would change that.

agronholm:
Perhaps I misunderstood the problem. I may need to read through it again.

agronholm:
Ahh...yes, after actually reading through the PR I'm starting to get it now. Sorry for the noise – a proper review will follow.

agronholm (approved):
Looks great, thanks!

PR #1139, opened 2026-09-18
Reject bool for interval trigger fields and as_timedelta

Pitchfork-and-Torch:
## Summary

`bool` subclasses `int`, so:

- `IntervalTrigger(seconds=True)` silently became a 1-second interval
- `CalendarIntervalTrigger(days=True)` silently became a 1-day interval
- `as_timedelta(True)` (used for misfire grace, lease duration, etc.) silently became 1 second

Raise `TypeError` for bool on these numeric fields instead of coercing.

## Test plan

- [x] Added `test_rejects_bool_interval_fields`, `test_as_timedelta_rejects_bool`, `test_rejects_bool_calendar_fields`
- [x] Verified locally: new tests pass


agronholm:
Explain why this is a practical problem.

Pitchfork-and-Torch:
bool is a subclass of int. IntervalTrigger(seconds=True) is accepted today and becomes a 1-second interval. CalendarIntervalTrigger(days=True) becomes 1 day. as_timedelta(True) becomes 1 second. That shows up when a flag leaks from config, env, or JSON into a numeric field. The PR rejects those before they schedule.

mattewid:
> That shows up when a flag leaks from config, env, or JSON into a numeric field

Wouldn't it be better to fix the potential configuration data leak?

I believe that if Python's `time(...)` and `timedelta(...)` functions accept these boolean values, then the "issue" doesn't lie with APScheduler. Another point is that you are using checks like `if isinstance(value, bool)`, but nothing prevents a user from passing a value of a different type; if an incorrect type is passed to `time(...)` or `timedelta(...)`, they would likely raise an exception.

agronholm:
Yeah, I asked why this is a *practical* problem and got an explanation why it's a *hypothetical* problem. I'm not sold at all on needing a bool check for this.

PR #1140, opened 2026-09-19
Reject trailing characters in special cron day expressions

jakezwang:

## Changes

`CronTrigger(day="last/2")` currently accepts the expression as `last`, and
`day="1st mon/2"` is accepted as `1st mon`. Anchor the two special day expressions
so trailing characters raise `ValueError`. Valid expressions, including mixed-case
names, continue to produce the same dates.

Validation: five regressions fail before the fix; all 399 local tests pass
(`pytest -m "not external_service"`, two skips), and all normal pre-commit hooks pass.
Independent review also verified 34 public API controls. Manual mypy reports the
same 95 errors on pristine master. Strict docs reproduce an existing
`TypeAliasForwardRef` warning; the docs build passes when that known warning is
excluded.

## Checklist

- [x] You've updated the documentation (in `docs/`, in case of behavior changes or new
features)

these instructions.

### Updating the changelog

```
```


docs/userguide.rst
```diff
@@ -382,6 +382,10 @@ number of built-in trigger classes:
   use when you want to run the task on calendar-based intervals, at a specific time of
   day
 
+Special ``day`` expressions in :class:`~triggers.cron.CronTrigger`, such as ``last``
+and ``1st mon``, must match in full. Trailing characters, such as ``/2`` in
+``last/2``, raise :exc:`ValueError` instead of being ignored.
+
```
agronholm, line 388: Historical notes belong in the changelog, not here.

```suggestion
```

tests/triggers/test_cron.py
```diff
@@ -13,6 +13,25 @@ def test_invalid_expression():
...
+)
+def test_special_day_expression(expr, expected_day, utc_timezone):
+    trigger = CronTrigger(
+        day=expr,
+        start_time=datetime(2026, 1, 1, tzinfo=utc_timezone),
+        timezone=utc_timezone,
+    )
+    assert trigger.next() == datetime(2026, 1, expected_day, tzinfo=utc_timezone)
+
+
```
agronholm, line 34: What does this test add to the fix?

agronholm (changes requested):
Why is this focused only on trailing characters and not leading characters too? Seems like that should be rolled into this, no?

agronholm:
I also question who in their right mind would think such an expression would even result in something reasonable. Was this PR a result of a practical problem?

jakezwang:
This was found by independent edge-case validation, not an observed production problem. I added leading-character cases alongside the trailing-character cases: leading junk is already rejected by `re.match`, while the missing end anchors allowed trailing junk. I removed the redundant positive cases and the historical user-guide paragraph, leaving the changelog entry. All 192 trigger tests pass locally. The latest CI failures are in scheduler/asyncpg state handling and a Windows aiosqlite timeout, not the cron tests. I attempted to rerun the failed jobs, but GitHub requires repository admin rights; a maintainer rerun would help determine whether those are transient.

tests/triggers/test_cron.py
```diff
@@ -13,6 +13,26 @@ def test_invalid_expression():
...
+        "lastx",
+        "last!",
+        "last/2",
+        "1st mon/2",
+        "last fri!",
+        "xlast",
+        "!last",
+        "2/last",
+        "x1st mon",
+        "!last fri",
```
agronholm: Wouldn't two parameter values suffice? I don't see what all these other values contribute.

